PR #2, opened 2017-07-31
run travis integration tests on scylla

mmatczuk:


penberg:
Why not run on *both* Apache Cassandra and ScyllaDB?

mmatczuk:
Sure can do that, are the Scylla flags OK?

penberg:
@mmatczuk Why do you need `--enable-in-memory-data-store`? Otherwise looks OK to me.

mmatczuk:
Not to write on disk during the test

penberg:
@mmatczuk Not write to disk to make the tests run faster? I don't object to using it, but I doubt it will show any significant difference. Up to you.

mmatczuk:
It's not the point, it runs in a Travis docker and we try to be nice to Travis by eliminating the writes.

mmatczuk:
@penberg please take a look

penberg:
:+1:

PR #9, opened 2017-08-23
notice update

mmatczuk:


penberg:
I haven't looked at the detailed requirements from ASF so let's CC @avikivity.

Please do drop the copyright symbol from the beginning of the trademark notice. It's just weird.

Also, typical format for copyright statement is:

```
Copyright (C) <year> <author>
```

You also need to explicitly state the license gocqlx is distributed under. For example,

```
Copyright (C) 2017 ScyllaDB

gocqlx is distributed under the Apache 2.0 license. See the LICENSE file for details.
```

mmatczuk:
@penberg updated accordingly

PR #21, opened 2017-10-25
Create simple documentation to help new contributors

danielsoro:


mmatczuk:
Thanks for the PR, I believe, however that instructing users how to install `moreutils` adds little value. We need a document on contributing, maybe it should contain information that `moreutils` is needed to run `check`.

Issue #19, opened 2017-10-03
qb feature request: Literal value specification in "set", "where" and "if"


jgiles:
In cases where queries are setting or comparing against some constant value (e.g. NULL, 0, "", a particular value of an enum-style text field), it would improve usability and readability to specify that constant value as a literal during query construction rather than making it a query parameter and then binding it later.

A natural API might be to add "Lit" variants in the same way there are currently "Named" variants:
qb.EqLit(column string, value string)
qb.GtLit(column string, value string)
qb.InLit(column string, valueSet string...)
...
and probably a .SetLit(column string, value string)

I've made the inputs strings here to avoid the question of value marshaling - if the input really is constant or constant-like, writing out the string should not be too burdensome for clients.

Or, if that adds unacceptable bloat, there could be some
qb.CmpLit(expression string)
where the caller is responsible for including the operator. That reduces API size and may be the most flexible, but it opens a fairly inviting hole for misuse in an otherwise pretty prescriptive API.

I'm happy to send a PR implementing this (and almost just sent one), but figured it would be good to discuss the desired API first.

mmatczuk:
Hi @jgiles,
thanks for the proposal! I like the "Lit" idea because it improves readability so we can do it.

jgiles:
So, a 'Lit' suffix on on the Eq, Gt, etc?

mmatczuk:
Yes.

jgiles:
Sweet, I'll send a PR (possibly in a couple days, I have a few other things in the queue at the moment).
Thanks!

mmatczuk:
Sure, no worries. Thanks!

mmatczuk:
Hi @jgiles how are you doing? I'm thinking of taking over this task, I do not want to get in your way.

jgiles:
@mmatczuk "in a couple days" - famous last words. I should have a PR up today, apologies for the delayed response.

jgiles:
In fact, I went just went ahead and put up https://github.com/scylladb/gocqlx/pull/22 which does the comparison changes.

jgiles:
#22 is now a more comprehensive fix, let me know what you think @mmatczuk

mmatczuk:
Closed in #37.

PR #22, opened 2017-10-29
Start #19 - support comparing against literals.

jgiles:
This adds *Lit variants of the Eq, Gt, Lt, etc comparison methods,
permitting convenient comparison against known constants.

Restructure Cmp.writeCql to avoid appends + named-parameter return for
clarity, which also reduces BenchmarkCmp-8 from 612 ns/op to 555 ns/op.

### DIFF
--- a/qb/insert.go
+++ b/qb/insert.go
@@ -11,10 +11,16 @@ import (
 	"bytes"
 )
 
+// initializer specifies an value for a column in an insert operation.
+type initializer struct {
+	column string
+	value  value
+}
+
 // InsertBuilder builds CQL INSERT statements.
 type InsertBuilder struct {
 	table   string
-	columns columns
+	columns []initializer
 	unique  bool
 	using   using
 }
@@ -37,12 +43,22 @@ func (b *InsertBuilder) ToCql() (stmt string, names []string) {
 	cql.WriteByte(' ')
 
 	cql.WriteByte('(')
-	b.columns.writeCql(&cql)
-	names = append(names, b.columns...)
+	nCols := len(b.columns)
+	for i, init := range b.columns {
+		cql.WriteString(init.column)
+		if i < nCols-1 {
+			cql.WriteByte(',')
+		}
+	}
 	cql.WriteString(") ")
 
 	cql.WriteString("VALUES (")
-	placeholders(&cql, len(b.columns))
+	for i, init := range b.columns {
+		names = append(names, init.value.writeCql(&cql)...)
+		if i < nCols-1 {
+			cql.WriteByte(',')
+		}
+	}
 	cql.WriteString(") ")
 
 	names = append(names, b.using.writeCql(&cql)...)
@@ -63,7 +79,39 @@ func (b *InsertBuilder) Into(table string) *InsertBuilder {
 
 // Columns adds insert columns to the query.
 func (b *InsertBuilder) Columns(columns ...string) *InsertBuilder {
-	b.columns = append(b.columns, columns...)
+	for _, c := range columns {
+		b.columns = append(b.columns, initializer{
+			column: c,
+			value:  param(c),
+		})
+	}
+	return b
+}
+
+// NamedColumn adds an insert column with a custom parameter name.
+func (b *InsertBuilder) NamedColumn(column, name string) *InsertBuilder {
+	b.columns = append(b.columns, initializer{
+		column: column,
+		value:  param(name),
+	})
+	return b
+}
+
+// LitColumn adds an insert column with a literal value to the query.
+func (b *InsertBuilder) LitColumn(column, literal string) *InsertBuilder {
+	b.columns = append(b.columns, initializer{
+		column: column,
+		value:  lit(literal),
+	})
+	return b
+}
+
+// FuncColumn adds an insert column initialized by evaluating a CQL function.
+func (b *InsertBuilder) FuncColumn(column string, fn *Func) *InsertBuilder {
+	b.columns = append(b.columns, initializer{
+		column: column,
+		value:  fn,
+	})
 	return b
 }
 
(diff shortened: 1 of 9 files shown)


jgiles:
The last commit is a slightly more aggressive refactor/expansion. I'm leaving it as a separate commit in case it's not appealing, but can squash all these if you like.

jgiles:
@mmatczuk thoughts on this?

mmatczuk:
Hi,
it looks good, I like the idea with value. I'm on holidays and will take a closer look next week.

qb/insert.go
```diff
@@ -63,7 +79,39 @@ func (b *InsertBuilder) Into(table string) *InsertBuilder {
...
+		b.columns = append(b.columns, initializer{
+			column: c,
+			value:  param(c),
+		})
+	}
+	return b
+}
+
+// NamedColumn adds an insert column with a custom parameter name.
+func (b *InsertBuilder) NamedColumn(column, name string) *InsertBuilder {
```
mmatczuk: By the convention it should be `ColumnNamed` right?

qb/insert.go
```diff
@@ -63,7 +79,39 @@ func (b *InsertBuilder) Into(table string) *InsertBuilder {
...
+	b.columns = append(b.columns, initializer{
+		column: column,
+		value:  param(name),
+	})
+	return b
+}
+
+// LitColumn adds an insert column with a literal value to the query.
+func (b *InsertBuilder) LitColumn(column, literal string) *InsertBuilder {
```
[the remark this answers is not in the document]
mmatczuk: So here maybe it should also be `ColumnLit`?

qb/insert.go
```diff
@@ -63,7 +79,39 @@ func (b *InsertBuilder) Into(table string) *InsertBuilder {
...
+	b.columns = append(b.columns, initializer{
+		column: column,
+		value:  lit(literal),
+	})
+	return b
+}
+
+// FuncColumn adds an insert column initialized by evaluating a CQL function.
+func (b *InsertBuilder) FuncColumn(column string, fn *Func) *InsertBuilder {
```
mmatczuk: And here `ColumnFunc`?

qb/insert_test.go
```diff
@@ -35,6 +35,18 @@ func TestInsertBuilder(t *testing.T) {
 			S: "INSERT INTO cycling.cyclist_name (id,user_uuid,firstname,stars) VALUES (?,?,?,?) ",
 			N: []string{"id", "user_uuid", "firstname", "stars"},
 		},
+		// Add a named column
+		{
+			B: Insert("cycling.cyclist_name").Columns("id", "user_uuid", "firstname").NamedColumn("stars", "stars_name"),
+			S: "INSERT INTO cycling.cyclist_name (id,user_uuid,firstname,stars) VALUES (?,?,?,?) ",
+			N: []string{"id", "user_uuid", "firstname", "stars_name"},
+		},
+		// Add a literal column
```
mmatczuk: Add test for `ColumnFunc`.

qb/update.go
```diff
@@ -9,31 +9,20 @@ package qb
...
 )
 
 // assignment specifies an assignment in a set operation.
 type assignment struct {
-	column string
-	name   string
-	expr   bool
-	fn     *Func
+	column      string
+	value       value
```
mmatczuk: :+1:

qb/update.go
```diff
@@ -106,47 +95,82 @@ func (b *UpdateBuilder) Set(columns ...string) *UpdateBuilder {
...
 }
 
 // AddNamed adds SET column=column+? clauses to the query with a custom
 // parameter name.
 func (b *UpdateBuilder) AddNamed(column, name string) *UpdateBuilder {
+	return b.addValue(column, param(name))
+}
+
+// AddLit adds SET column=column+literal clauses to the query.
+func (b *UpdateBuilder) AddLit(column, literal string) *UpdateBuilder {
```
mmatczuk: `lit interface{}`

qb/update.go
```diff
@@ -106,47 +95,82 @@ func (b *UpdateBuilder) Set(columns ...string) *UpdateBuilder {
...
+}
+
+func (b *UpdateBuilder) addValue(column string, value value) *UpdateBuilder {
 	b.assignments = append(b.assignments, assignment{
-		column: fmt.Sprint(column, "=", column, "+?"),
-		name:   name,
-		expr:   true,
+		column:      column,
+		value:       value,
+		valuePrefix: column + "+",
```
mmatczuk: Instead of that maybe add and remove should be values too?

qb/value.go
```diff
@@ -0,0 +1,37 @@
+package qb
```
mmatczuk: File header is missing.

mmatczuk: Maybe migrate `func` file to `value`?

qb/value.go
```diff
@@ -0,0 +1,37 @@
...
+
+import "bytes"
+
+// value is a CQL value expression for use in an initializer, assignment, or comparison.
+type value interface {
+	// clone produces a clone of this value. Writes to the original and the clone are independent.
+	// It also marks a type as implementing the value interface.
+	clone() value
```
mmatczuk: Clone method is not used and not really needed, please remove it from this PR.

jgiles:
I'm sorry I disappeared on this, I've been pulled into a lot of other things.

In the immediate term I'm abroad on vacation for the next week. If someone wants to pick this up and finish it off I am not opposed.

mmatczuk:
@jgiles thanks for your good work on this. @annismckenzie I'll handle this tomorrow or early next week.

mmatczuk:
@jgiles work moved to #37

PR #24, opened 2017-10-29
Add ColumnAs for 'SELECT col AS name' queries.

jgiles:
Clients could do this hackily before by manually passing "col AS name",
but it isn't clearly supported.

### DIFF
--- a/qb/select.go
+++ b/qb/select.go
@@ -115,10 +115,9 @@ func (b *SelectBuilder) Columns(columns ...string) *SelectBuilder {
 	return b
 }
 
-// ColumnAs adds a result column with a custom name to the query.
-func (b *SelectBuilder) ColumnAs(column, name string) *SelectBuilder {
-	b.columns = append(b.columns, column+" AS "+name)
-	return b
+// As is a helper for adding a column AS name result column to the query.
+func As(column, name string) string {
+	return column + " AS " + name
 }
 
 // Distinct sets DISTINCT clause on the query.
--- a/qb/select.go
+++ b/qb/select.go
@@ -115,6 +115,12 @@ func (b *SelectBuilder) Columns(columns ...string) *SelectBuilder {
 	return b
 }
 
+// ColumnAs adds a result column with a custom name to the query.
+func (b *SelectBuilder) ColumnAs(column, name string) *SelectBuilder {
+	b.columns = append(b.columns, column+" AS "+name)
+	return b
+}
+
 // Distinct sets DISTINCT clause on the query.
 func (b *SelectBuilder) Distinct(columns ...string) *SelectBuilder {
 	b.distinct = append(b.distinct, columns...)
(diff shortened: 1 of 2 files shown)


mmatczuk:
My objection is that the submitted API is hard to use, consider that you have the columns `a`, `b`, `c` and `d` and you want to have `b` as `d` then you have to split the columns setting into 2 or 3 calls.

```go
// with the patch
qb.Select("test").Columns("a", "c", "d").ColumnAs("b", "d")
// without the patch
qb.Select("test").Columns("a", "b AS d", "c", "d")
// other possibility
qb.Select("test").Columns("a", qb.As("b, "d"), "c", "d")
```

Adding `qb.As(s, name string)` function seems more flexible.

qb/select_test.go
```diff
@@ -28,6 +28,11 @@ func TestSelectBuilder(t *testing.T) {
 			B: Select("cycling.cyclist_name").Columns("id", "user_uuid", "firstname"),
 			S: "SELECT id,user_uuid,firstname FROM cycling.cyclist_name ",
 		},
+		// Add a SELECT AS column.
```
mmatczuk: Please follow the convention and do not put a dot at the end here. It's not a sentence anyway.

jgiles:
re: `qb.As("b", "d")` vs `.ColumnAs("b", "d")` the thought was to mirror the conceptually-similar `Named` update/insert variants, which also force you to split your specification into multiple lines. However, it looks like there isn't the `(*updateBuilder) setNamed(column, name string)` I had thought there was, so maybe that precedent doesn't exist.

The only other trouble is that `qb.As` commits qb to supporting the 'b AS d' string literal insertion implementation for `AS` support. But if you don't anticipate that being a problem, `qb.As` is definitely more convenient for clients.

mmatczuk:
> The only other trouble is that qb.As commits qb to supporting the 'b AS d' string literal insertion implementation for AS support

I'm not sure I understand it, could you please explain?

jgiles:
You could imagine keeping track of the various aliases similar the to tracking of names, and making them inspectable/using them for some kind of validation. But I don't have any concrete proposals, so I don't think it matters.

My reluctance is probably just lingering guilt/doubt about codifying the hacky-feeling 'foo AS bar' code I've already written into the library itself :-) if you're on board I can let those doubts go.

I'll go ahead and make the change.

qb/select_test.go
jgiles: Done

qb/select.go
```diff
@@ -115,10 +115,9 @@ func (b *SelectBuilder) Columns(columns ...string) *SelectBuilder {
 	return b
 }
 
-// ColumnAs adds a result column with a custom name to the query.
-func (b *SelectBuilder) ColumnAs(column, name string) *SelectBuilder {
-	b.columns = append(b.columns, column+" AS "+name)
-	return b
+// As is a helper for adding a column AS name result column to the query.
```
mmatczuk: Drop the previous commit, move the function it to `utils.go` and add a smaller unit test there.

mmatczuk:
Squashed and merged, thanks!

PR #26, opened 2017-11-06
Do not release queries in Get/Select. Fix #25.

### LINKED ISSUE #25: Get/Select SHOULD NOT release queries
Currently, the `Get` and `Select` iterx methods call `iter.ReleaseQuery()`. This is a highly dangerous "convenience" which can lead to data races under common usage patterns.

The `Iter` constructor and `Get` and `Select` functions accept a query as a parameter, and so from  an API design perspective it make most sense to let clients (who created the query) manage the query lifecycle.

E.g. it's a pretty common pattern to do
```
q := createQuery()
defer q.Release()
// do stuff with query, err returns, etc
return result, nil
```

But if "doing stuff" includes a `gocqlx.Select`, this pattern will lead to `Release` being called twice. That means the query object now has two entries in the query pool, which will distribute it to two separate goroutines, and racy concurrent read/writes ensue. We had a very hard-to-track-down bug in our service caused by this.

In this case, the cost of *not* calling Release is minimal; the query will just be GC'd instead of recycled. It's safer and better to let clients do releasing themselves.

jgiles:
Releasing query objects in Get/Select could easily lead to
double-releases, which can cause dangerous and tricky data races.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -146,6 +146,7 @@ func TestExample(t *testing.T) {
 		q := gocqlx.Query(session.Query(stmt), names).BindMap(qb.M{
 			"first_name": "Patricia",
 		})
+		defer q.Release()
 
 		var p Person
 		if err := gocqlx.Get(&p, q.Query); err != nil {
@@ -165,6 +166,7 @@ func TestExample(t *testing.T) {
 		q := gocqlx.Query(session.Query(stmt), names).BindMap(qb.M{
 			"first_name": []string{"Patricia", "Igy", "Ian"},
 		})
+		defer q.Release()
 
 		var people []Person
 		if err := gocqlx.Select(&people, q.Query); err != nil {
@@ -190,6 +192,7 @@ func TestExample(t *testing.T) {
 			ToCql()
 
 		q := gocqlx.Query(session.Query(stmt), names).BindStruct(p)
+		defer q.Release()
 
 		var people []Person
 		if err := gocqlx.Select(&people, q.Query); err != nil {
--- a/iterx.go
+++ b/iterx.go
@@ -26,8 +26,7 @@ func Select(dest interface{}, q *gocql.Query) error {
 // Iterx is a wrapper around gocql.Iter which adds struct scanning capabilities.
 type Iterx struct {
 	*gocql.Iter
-	query *gocql.Query
-	err   error
+	err error
 
 	unsafe bool
 	Mapper *reflectx.Mapper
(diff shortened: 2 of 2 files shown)


jgiles:
It would probably also make sense to remove ReleaseQuery entirely (it's hard to explain how it should fit into the query lifecycle). But, that's a breaking API change.

mmatczuk:
> It would probably also make sense to remove ReleaseQuery entirely (it's hard to explain how it should fit into the query lifecycle). But, that's a breaking API change.

`ReleaseQuery` could easily be made private no need to export it.

> Releasing query objects in Get/Select could easily lead to double-releases, which can cause dangerous and tricky data races.

That's true, but I'd rather keep the behavior not to break existing code.

We could:

* introduce `gocqlx.AutoRelease` flag set to `true` by default
* introduce `Iterx.release` flag and a setter for it `Release(release bool)`
* release only when the flag is true
* rename `ReleaseQuery` to `releaseQuery`

WDYT?

jgiles:
> `ReleaseQuery` could easily be made private no need to export it.

It's possible there are clients who are currently calling the method, but if we don't care about breaking them we can remove it.

> That's true, but I'd rather keep the behavior not to break existing code.

Removing the release isn't going break anyone's code. At worst they will face a minor performance regression in the form of slightly increased GC pressure.

On the other hand, continuing to do the release is "dangerous by default" - it has caused and will continue to cause application correctness bugs. Bugs of the worst kind - sporadic, non-reproducible, and difficult to track down even with the Go race detector (because the races don't directly involve the read/write operations in the ReleaseQuery call, instead cropping up between reads/writes of two later goroutines when the double-released objects are re-issued from the `sync.Pool`).

And on the performance regression... recently I've begun to doubt whether the gocql query pooling is even worth it. The number unpooled of allocations involved in actual query binding/execution/scanning may dwarf the benefits of pooling the query objects. At some point I want to put together a benchmark... gocql doesn't seem to have one, which itself is a red flag for a performance optimization feature.

jgiles:
Agreed on removing the query field of iterx entirely; you can't reliably check against use of already-released queries anyway (client could release it directly).

I've added a commit that does this. It's a breaking API change because the ReleaseQuery method is removed. If that makes sense, I can squash these.

mmatczuk:
Please squash it, also if you could update examples and readme with defer release on query it would be great!

jgiles:
Squashed and updated the docs.

Even the `ExecRelease` utility is beginning to look a little dangerous to me, but at least it is very explicit about what it does.

mmatczuk:
Thanks!

PR #30, opened 2017-12-06
Create migrator table interface and add tests for the migrations

bilias23:
The `migrate` functionality currently has no tests. This makes it harder to refactor and add further functionality. 
This PR implements an abstraction for the migrations table to be used while testing and adds some basic tests.
It also adds two minor behavioral changes: 

1. Providing feedback if no migration files were found (or directory was not valid)
2. Providing feedback if no CQL statements were found in a migration file.

The behavioral changes could also be split into a separate MR if desired. 

I'd also like to thank you for this library and hope you accept contributions. :)


mmatczuk:
@bilias23 @annismckenzie thank you for the patch. As it's a big change I'll need some time to evaluate it.

migrate/migrate.go
```diff
@@ -122,6 +122,7 @@ func applyMigration(ctx context.Context, mt MigratorTable, path string, done int
 	}
 
 	i := 1
+	stmtcount := 0
```
mmatczuk: Could you rename it to `stmtCount`?

annismckenzie:
Sorry for being unresponsive. Sad that you don't like, though. I agree on the integration test part and will prepare something. As we already have another feature in our fork (adding the ability to read the migration files from an embedded `http.FileSystem`) and because that is based on this work, I'm gonna have to maintain our fork for a longer time.

I'll send those two commits in a separate PR later this week.

mmatczuk:
> Sorry for being unresponsive

No worries.

> As we already have another feature in our fork (adding the ability to read the migration files from an embedded http.FileSystem) 

What could be an interesting contribution is adding

```go
func Migrate(ctx context.Context, session *gocql.Session, p Provider) error
``` 

based on abstract `migrate.Provider`

```go
type Provider interface {
List() ([]string, error)
Open(name string) (os.ReadCloser, error)
}
```

WDYT?

annismckenzie:
Well, both `dir string` (via `http.Dir`) as well as any other virtual file system implement `http.FileSystem`, making the current method signature backwards-compatible (`dir string` is just wrapped in `http.Dir()`). I currently have a fork-only PR open for a colleague and me to collaborate and comment on [here](https://github.com/alfatraining/gocqlx/pull/3). The `migrate.Provider` interface would just be another layer of abstraction for what we're doing over there, you'd still end up with `NewNativeFileSystemProvider(dir string) *osFileProvider` that would wrap `http.Dir` and coax the return values to your interface methods. I mean, yeah, `http.FileSystem` is heavy… now I'm right back at the beginning.

mmatczuk:
Using http interfaces introduces dependency to http that is hard to justify. The package shall define it's own interfaces that are convenient to use and reflect the domain...

Unit testing this must be strange as http.File inherits from so many interfaces that are irrelevant.

mmatczuk:
I'm closing this as #49 implements testing for migrations and #48 adds capability to execute custom code around migrations.

PR #31, opened 2017-12-14
Avoid one allocation when binding structs

nussjustin:
This speeds up binding structs by using the new TraversalsByNameFunc introduced in jmoiron/sqlx#358, which avoids allocating a slice.

```
name          old time/op    new time/op    delta
BindStruct-8     519ns ± 2%     400ns ± 2%  -22.97%  (p=0.000 n=10+10)

name          old alloc/op   new alloc/op   delta
BindStruct-8      224B ± 0%      128B ± 0%  -42.86%  (p=0.000 n=10+10)

name          old allocs/op  new allocs/op  delta
BindStruct-8      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.000 n=10+10)
```

### DIFF
--- a/queryx.go
+++ b/queryx.go
@@ -134,21 +134,22 @@ func bindStructArgs(names []string, arg0 interface{}, arg1 map[string]interface{
 		v = v.Elem()
 	}
 
-	fields := m.TraversalsByName(v.Type(), names)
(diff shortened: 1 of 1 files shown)


mmatczuk:
Looks good, thanks!

PR #32, opened 2017-12-18
Provide feedback if no files or statements found

bilias23:
As discussed on #30 this is a separate PR with the two commits that provide feedback if no migration files were found or if no CQL statements were found in a migration file.

Thanks :)

### DIFF
--- a/migrate/migrate.go
+++ b/migrate/migrate.go
@@ -21,6 +21,8 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// ErrNoMigrationsFound is returned if the directory provided doesn't contain any migrations.
+var ErrNoMigrationsFound = errors.New("no migrations were found")
 const (
 	infoSchema = `CREATE TABLE IF NOT EXISTS gocqlx_migrate (
 	name text,
@@ -78,6 +80,10 @@ func Migrate(ctx context.Context, session *gocql.Session, dir string) error {
 	if err != nil {
 		return fmt.Errorf("failed to list migrations in %q: %s", dir, err)
 	}
+	if len(fm) == 0 {
+		return ErrNoMigrationsFound
+	}
+
 	sort.Strings(fm)
 
 	// verify migrations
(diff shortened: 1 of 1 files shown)


migrate/migrate.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// ErrNoMigrationsFound is returned if the directory provided doesn't contain any migrations.
+var ErrNoMigrationsFound = errors.New("no migrations were found")
+
+// ErrNoCQLStmt is returned when trying to execute a non cql statement i.e. "This is not cql"
+var ErrNoCQLStmt = errors.New("no cql statement found")
```
mmatczuk: This error is a very rare case when you read empty files only. I suggest reusing ErrNoMigrationsFound.

migrate/migrate.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// ErrNoMigrationsFound is returned if the directory provided doesn't contain any migrations.
+var ErrNoMigrationsFound = errors.New("no migrations were found")
```
mmatczuk: How about renaming to `ErrNoMigrations`

mmatczuk: Why do you think this should be en exported error variable?

migrate/migrate.go
bilias23: It could also be triggered if a file does not contain ";". In any case it is a rare occurrence, so I agree to use the same: errNoMigrations

migrate/migrate.go
```diff
@@ -21,6 +21,9 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// ErrNoMigrations is returned if the directory provided doesn't contain any migrations, or if a migration file does not contain a CQL statement
+var ErrNoMigrations = errors.New("no migrations were found")
```
mmatczuk: Why do you thinks this error shall be exported, are there any other errors that should be exported too?

bilias23: Actually no, it is a remnant of #30. It does not need to be exported (at least for now).

mmatczuk: OK so if it's not exported maybe it should be inlined WDYT?

bilias23: You mean like the rest of the errors like so?
`return fmt.Errorf("no migrations found in %q", dir)`

Sure that could work, we can change it if and when we have to.

mmatczuk:
Thanks!

PR #35, opened 2017-12-27
Fix query order when using ttl and unique

ericrenard:
Correctly builds a query when using TTL and Unique. Current version generates invalid CQL statement.

Example:
qb.Insert("table").Columns("a", "b", "c").Unique().TTL().ToCql()

Produces:
INSERT INTO table (a,b,c) VALUES (?,?,?) USING TTL ? IF NOT EXISTS

Should produce:
INSERT INTO table (a,b,c) VALUES (?,?,?) IF NOT EXISTS USING TTL ?


mmatczuk:
Thanks!

PR #37, opened 2018-04-20
Support for literals in INSERT and UPDATE and comparisons

mmatczuk:
The 'value' interface represents a CQL value for use in a comparison,
update, or intitialization operation. A consistent interface for this
allows us to easily support specifying default-named, custom-named,
literal, and evaluated-function values in all these contexts.

Parameters to Func should probably also be values to support full
composition, but that would be a breaking change because Func's
properties are exposed.

The value interface could itself be exposed if we wanted to allow
clients to pass their own values to SetValue, etc, but for now it is a
package-internal abstraction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,17 @@ hood it uses `sqlx/reflectx` package so `sqlx` models will also work with `gocql
 
 ## Features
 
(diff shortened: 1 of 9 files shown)


mmatczuk:
Thanks @jgiles !!

PR #38, opened 2018-05-09
Document data type mappings

josephtaylor:
Took me a little bit to find the mappings in the integration test (was looking specifically for decimal).

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,28 @@ type Person struct {
 
 See more examples in [example_test.go](https://github.com/scylladb/gocqlx/blob/master/example_test.go).
 
+## Data-Type Mappings
+
+| cassandra               | go structs          |
+|-------------------------|---------------------|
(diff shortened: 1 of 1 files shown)


mmatczuk:
Hi guys, thanks for the contribution. All this actually relates to gocql the driver but I'd be happy to accept the donation. There is an inconsistency, however, with list and maps, what with `list<int>`? Another thing that would be good to add is UDT and custom marshallers. Let me know WDYT.

mmatczuk:
My suggestion would be to eliminate collections from the table and somewhat describe it underneath.

josephtaylor:
Yeah that makes sense cause the collection ones are more examples than definitions

josephtaylor:
Going back to your original comment, I think you're right! Since this is really describing the functionality of the gocql driver, maybe this isn't the best place for it to be documented? :man_shrugging:

PR #39, opened 2018-05-11
Add aggregation function to select query builder

SpiritOfWill:


qb/func.go
```diff
@@ -49,3 +49,12 @@ func MaxTimeuuid(name string) *Func {
 func Now() *Func {
 	return Fn("now")
 }
+
+// Count produces COUNT(field_name) or COUNT(*).
+func Count(paramNames ...string) *Func {
+	name := "COUNT"
```
mmatczuk: In other places we use lowercase function names, can we be consistent on that?

qb/func.go
```diff
@@ -49,3 +49,12 @@ func MaxTimeuuid(name string) *Func {
...
+	if len(paramNames) > 0 {
+		return Fn(name, paramNames...)
+	}
+	return Fn(name, "*")
+}
```
mmatczuk: How about adding Min and Max? They Seem to be frequently used functions too.

qb/select.go
```diff
@@ -163,3 +163,14 @@ func (b *SelectBuilder) AllowFiltering() *SelectBuilder {
...
 }
+
+// Func sets aggregation function for select clause on the query.
+func (b *SelectBuilder) Func(fn *Func) *SelectBuilder {
+	var param string
+	if len(fn.ParamNames) > 0 {
+		param = fn.ParamNames[0]
+	}
+
+	b.Columns(fn.Name + "(" + param + ")")
```
mmatczuk: After thinking it over it appears to me that we should have the very same thing as for insert builder here, https://github.com/scylladb/gocqlx/blob/master/qb/insert.go#L98-L114. 

`ColumnFunc` would let you call server side functions with bounded params, ColumnLit would add any literal the way you want. On top literals we can add `Count`.

Anyway, why not just use just `Columns`?

```
	B: Select("cycling.cyclist_name").Columns("MAX(stars) as max_stars").GroupBy("id"),
	S: "SELECT id,MAX(stars) as max_stars FROM cycling.cyclist_name GROUP BY id ",
```

qb/func.go
SpiritOfWill: OK.

qb/func.go
SpiritOfWill: OK.

qb/select.go
SpiritOfWill: 1. OK.
2. `Anyway, why not just use just Columns?` - I'm going to use `count(*)` quite often, so I'd prefer to have the method for it.

mmatczuk: @SpiritOfWill not to overdo it, how about just adding aggregate functions [1] to SelectBuilder. For example `Count` function that would append columns with appropriate cql code as LitColumn does [2]. 

WDYT?

[1] https://docs.datastax.com/en/cql/3.3/cql/cql_using/useQueryStdAggregate.html
[2] https://github.com/scylladb/gocqlx/blob/master/qb/insert.go#L99

SpiritOfWill: @mmatczuk OK. I'll just add `Count` method to `SelectBuilder`?

mmatczuk: For completeness it would be better to add all of them: min, max, avg, sum, and count. They are pretty similar.

SpiritOfWill: @mmatczuk Done.

qb/select.go
```diff
@@ -163,3 +163,44 @@ func (b *SelectBuilder) AllowFiltering() *SelectBuilder {
 	b.allowFiltering = true
 	return b
 }
+
+// Func sets aggregation function for select clause on the query.
+func (b *SelectBuilder) Func(name, column string) *SelectBuilder {
```
mmatczuk: I'd not export that function, if you want sth custom you can use Columns directly. Func is sort of a name clash with Func struct.

Also can we put this function bellow the others?

qb/select.go
```diff
@@ -163,3 +163,44 @@ func (b *SelectBuilder) AllowFiltering() *SelectBuilder {
...
+	b.Columns(name + "(" + column + ")")
+	return b
+}
+
+// Count produces 'count(column)' or 'count(*)'.
+func (b *SelectBuilder) Count(column ...string) *SelectBuilder {
```
mmatczuk: How about explicitly making `Count(column string)` and `CountAll()` functions.

qb/select.go
```diff
@@ -163,3 +163,44 @@ func (b *SelectBuilder) AllowFiltering() *SelectBuilder {
 	b.allowFiltering = true
 	return b
 }
+
+// Func sets aggregation function for select clause on the query.
+func (b *SelectBuilder) Func(name, column string) *SelectBuilder {
+	b.Columns(name + "(" + column + ")")
```
mmatczuk: How about invoking append directly `b.columns = append(b.columns, name + "(" + column + ")")`, also `fmt.Sprint` would be better here.

qb/select.go
```diff
@@ -163,3 +163,44 @@ func (b *SelectBuilder) AllowFiltering() *SelectBuilder {
...
+	p := "*"
+	if len(column) > 0 {
+		p = column[0]
+	}
+
+	b.Func("count", p)
+	return b
+}
+
+// Min produces 'min(column)' aggregation function
```
mmatczuk: dot at the end

mmatczuk: Applies to all the bellow functions.

qb/select.go
SpiritOfWill: This will break encapsulation principle, cause `b.columns = append(b.columns, name + "(" + column + ")")` - is internal realization and `b.Columns()` - is external API that should be stable or bacward compatible...

SpiritOfWill: And I think, that with `b.Columns()` this will be more supportable.
But if you insist, I will change it to `append`.

qb/select.go
SpiritOfWill: Done.

qb/select.go
SpiritOfWill: Done.

qb/select.go
SpiritOfWill: Done.

mmatczuk:
Thanks @SpiritOfWill !!

PR #50, opened 2018-05-24
Added checkErrAndNotFound to Select func

soluchok:


mmatczuk:
If I might ask why?

soluchok:
We need to know that nothing found in the database. We did some check like checkErrAndNotFound, but it will be better add this to common lib.

mmatczuk:
This change is not backward compatible and would break code. The current approach is modeled after REST where (usually) empty list is a valid result. If that does not suit your needs you may always check if list is not empty.

Issue #46, opened 2018-05-23
Add go.mod file


dahankzter:
To make it useable with `vgo` as given in golang/go#24301 
This is a relatively contained project that has the potential to work nicely without much work.

PR #55, opened 2018-06-01
generated an initial go.mod file.

dahankzter:
Fixes: #46

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,6 @@
+module github.com/scylladb/gocqlx
+
+require (
+	github.com/gocql/gocql v0.0.0-20180530083731-3c37daec2f4d
+	github.com/google/go-cmp v0.2.0
+)


go.mod
```diff
@@ -0,0 +1,6 @@
+module github.com/scylladb/gocqlx
+
+require (
+	github.com/gocql/gocql v0.0.0-20180530083731-3c37daec2f4d
+	github.com/google/go-cmp v0.2.0
```
mmatczuk: This is a test requirement do we have some distinction for that in vgo?

dahankzter: Afaik no. I think it works the same way as today. The test deps are just ignored bu the build commands and just used when actually running tests.

dahankzter: If they are not listed here the tests won't work however so I guess they are needed in the mod file.

PR #57, opened 2018-07-27
Add qb.ContainsKey to query keys in a map

vrischmann:
This is needed to query the keys of a map using a secondary index like this:

    CREATE TABLE user (id uuid primary key, cities map<text, text>);
    CREATE INDEX foobar ON user (keys(cities));
    SELECT * FROM user WHERE cities CONTAINS KEY 'Paris'

### DIFF
--- a/qb/cmp.go
+++ b/qb/cmp.go
@@ -19,6 +19,7 @@ const (
 	geq
 	in
 	cnt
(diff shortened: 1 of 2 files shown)


mmatczuk:
Thanks!

PR #58, opened 2018-07-28
makefile: switched to golangci-lint

mmatczuk:


Makefile
```diff
@@ -44,19 +53,13 @@ bench:
...
 get-tools: GOPATH := $(shell mktemp -d)
 get-tools:
 	@echo "==> Installing tools at $(GOBIN)..."
 	@mkdir -p $(GOBIN)
 
-	@go get -u github.com/golang/lint/golint
 	@go get -u github.com/client9/misspell/cmd/misspell
-	@go get -u github.com/gordonklaus/ineffassign
-	@go get -u honnef.co/go/tools/cmd/megacheck
+	@go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
```
jirfag: For CI I recommend installing binary version by curl to fix the version of linter

mmatczuk: Well spotted, is there a way to install the latest version automatically?

jirfag: Yes, but it has the same problem as go get-ing: one update of golangci-lint can cause failures of all your CI builds

mmatczuk: Done, I tried using the service but cannot access org public repos...

Issue #4, opened 2017-08-01
Add a code generator that would extract tables and columns from a database


mmatczuk:
## Flags
We need pretty much all the flags from https://github.com/volatiletech/sqlboiler#initial-generation, schema should be replaced with keyspace and models should be written to models/keyspace.

## Output
For every table in a given keyspace generate Table metatata.

```go
// TableName is generated metadata for <keyspace>.<table_name>.
struct TableName {
    Keyspace string
    Name string
    PK string[]
    Columns string[]
    ColumnName string
    ...
}
```

We shall follow the golang naming conventions and snake_case table / column names should be converted to camel case.

mmatczuk:
Cluster connectivity flags should be compatible with https://github.com/gocql/gocql/blob/master/common_test.go#L15.

SaubereSache:
I'd really appreciate an implementation of this as it removes a lot of boilerplate code

mmatczuk:
Any volunteers?

Bobochka:
Is it something like https://github.com/Bobochka/gocqlxschemagen that you looking for? If yes, how do you see the integration, should it be a separate bin inside this repo?

mmatczuk:
Nice work thanks.

Let's have a `cmd/schemagen` and put it there.

It would be nice to:
* put template to a separate file and embed it
* have an integration test 
* update readme
* provide example(s)

Bobochka:
@mmatczuk https://github.com/scylladb/gocqlx/pull/201

mmatczuk:
Fixed.

PR #63, opened 2018-09-13
Adding CreateTable to QueryBuilder

ymazdy:
Adding CreateTable to QueryBuilder for creation of tables from Structs based on gocqlx snake case based naming
it has been tested and used with ScyllaDB

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ type Person struct {
 
 /// DB Table Creation
 {
-    cql := qb.CreateTable("gocqlx_test.person", &Person{})
+    cql := qb.CreateTable("gocqlx_test.person").Columns(&Person{}).ToCql()
 	err := session.Query(cql).Exec()
     if err != nil {
         t.Fatal(err)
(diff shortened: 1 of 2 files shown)


mmatczuk:
This is an interesting POC, thanks for sharing! 

I think that it's somewhat similar to https://github.com/scylladb/gocqlx/issues/4, I'd rather have the code generation for model creation in a separate package. I have some doubts regarding production readiness of this code. Currently it supports only few types, there are questions about primary, sort keys and create table has much more features...

Overall I think it could make an interesting CLI tool that would consume go source file and produced some kind of cql.

ymazdy:
thanks for the comment, i used it for my own cluster i decided to share it

> This is an interesting POC, thanks for sharing!
> 
> I think that it's somewhat similar to #4, I'd rather have the code generation for model creation in a separate package. I have some doubts regarding production readiness of this code. Currently it supports only few types, there are questions about primary, sort keys and create table has much more features...
> 
> Overall I think it could make an interesting CLI tool that would consume go source file and produced some kind of cql.

thanks for the comment, i used it for my own cluster i decided to share it

Issue #69, opened 2018-10-24
Malformated query when there are only GroupBy


sagarafr:
For example, if we make a request with only a GroupBy statement
```go
package main

import (
	"github.com/sagarafr/gocqlx/qb"
	"fmt"
)

func main() {
	s, _ := qb.Select("cycling.cyclist_name").GroupBy("id").ToCql()
	fmt.Println(s)
}
```

we have this generation : 
```sql
SELECT id, FROM cycling.cyclist_name GROUP BY id
```

and this error, during the execution, because of the comma near FROM :
```cmdline
no viable alternative at input 'FROM'
```

PR #70, opened 2018-10-24
fix(qb): Fixed group by generation

sagarafr:
* In qb/select.go, if there are no colums and the query have a GroupBy
statement then the query will be malformated.
* In qb/utils.go, refactor writeCql with a strings.Join call.

[Issue #69](https://github.com/scylladb/gocqlx/issues/69)

### DIFF
--- a/qb/select.go
+++ b/qb/select.go
@@ -10,6 +10,7 @@ package qb
 import (
 	"bytes"
 	"fmt"
+	"strings"
 )
 
 // Order specifies sorting order.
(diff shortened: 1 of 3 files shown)


sagarafr:
I had one unit test for this case.
Your solution works, but I think my solution is "go-oriented". I run your benchmark on qb section, and it's the same.
But if you want I can put your solution :wink:

mmatczuk:
Hi @sagarafr,
thanks for the unit test, I must have missed your reply...
I'd rather avoid the string construction. Can you provide some benchmark data?

mmatczuk:
Thanks!

PR #71, opened 2018-10-26
 table: introduced table package

mmatczuk:
It adds support for super simple CRUD operations based on table schema model.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -15,7 +15,11 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
-var personSchema = `
+func TestExample(t *testing.T) {
+	session := CreateSession(t)
+	defer session.Close()
+
+	const personSchema = `
 CREATE TABLE IF NOT EXISTS gocqlx_test.person (
     first_name text,
     last_name text,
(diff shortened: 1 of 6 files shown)


table/table.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Select returns select by partition key statement.
+func (t *Table) Select(columns ...string) (stmt string, names []string) {
+	if len(columns) == 0 {
+		return t.sel.stmt, t.sel.names
+	}
+
+	return qb.Select(t.metadata.Name).
+		Columns(columns...).
+		Where(t.primaryKey[0:len(t.metadata.PartKey)]...).
```
rjeczalik: since it's used in two places how about making it a helper?

```go
func (t *Table) primaryKeyCmp() []qb.Cmp {
    return t.primaryKey[:len(t.metadata.PartKey)]
}
```

example_test.go
```diff
@@ -15,49 +15,48 @@ import (
...
-	session := CreateSession(t)
-	defer session.Close()
-
 	if err := ExecStmt(session, personSchema); err != nil {
 		t.Fatal("create table:", err)
 	}
 
+	// Person represents a row in person table.
+	// Field names are converted to camel case by default, no need to add special tags.
+	// If you want to disable a filed add `db:"-"` tag, it will not be persisted.
```
dahankzter: Typo `filed`.

table/table.go
mmatczuk: Fixed that with

```go
type Table struct {
	metadata      Metadata
	primaryKeyCmp []qb.Cmp
	partKeyCmp    []qb.Cmp
```

example_test.go
mmatczuk: Thanks!

PR #74, opened 2018-11-03
test(qb): Add tests for qb section

sagarafr:
* Add some missing test to get the 100% test coverage in the qb section


mmatczuk:
Ping @sagarafr, do we merge this work? I think we should but without the TTL thing as we discussed WDYT?

mmatczuk:
@sagarafr thanks, could you rebase it on top of master and squash the commits please.

sagarafr:
@mmatczuk Like this ?

PR #84, opened 2019-01-06
Update gocql version

rinx:
Due to gocql's issue (ex. https://github.com/gocql/gocql/issues/1191), the current version of gocqlx takes too high CPU utilization.
We updated go.mod file to use the newest version of gocql.

Thanks.


mmatczuk:
Thanks.

PR #90, opened 2019-02-01
Timestamps should be in milliseconds, not microseconds

VibyJocke:
Currently timestamps seem to be converted as microseconds, which is [incorrect](https://docs.scylladb.com/getting-started/types/#timestamps) and incompatible with other tools and client implementations (e.g. cqlsh).

This PR is just an illustrative example, but I guess you might want to fix it in a more backwards compatible way.


mmatczuk:
@VibyJocke if you are seeking to save as a timestamp then use and bind `time.Time` struct.

VibyJocke:
> @VibyJocke if you are seeking to save as a timestamp then use and bind `time.Time` struct.

Oh, thanks! I indeed thought it was the datatype `Timestamp` 🤦‍♂️

PR #91, opened 2019-02-13
fix gocql.Query.WithContext retrun shallow copy bug

kfrico:
fix gocql.Query.WithContext retrun shallow copy bug


mmatczuk:
Thanks!

PR #92, opened 2019-03-05
inherit the mapper of query

kingluo:
The temporary `Iter` should inherit the mapper from the query, otherwise it would hard-code as `DefaultMapper`, ignoring the actual mapper.


mmatczuk:
IMO this should be part of `Iter` function.

PR #93, opened 2019-03-05
inherit the mapper of query

kingluo:


mmatczuk:
I still think that the place to fix it is https://github.com/scylladb/gocqlx/pull/93/files#diff-5db579516f464ed50b8104cdafd2763aR47

kingluo:
change `Iter()` to accept `mapper`?

mmatczuk:
Fixed with #124

Issue #98, opened 2019-04-23
Support for CAS Operation Insight (LWT)


Linuturk:
https://docs.datastax.com/en/cql/3.3/cql/cql_using/useInsertLWT.html

I'm using the gocqlx library and performing a conditional update with the query builder. I'm not seeing an obvious way to determine whether a query was applied vs experiencing an error.

In plain gocql library, functions like https://godoc.org/github.com/gocql/gocql#Query.MapScanCAS return an applied bool to check whether the update was successful.

I don't see any similar return for Exec() or ExecRelease() using gocqlx. https://godoc.org/github.com/scylladb/gocqlx#Queryx.Exec Do these functions return an error when the update fails to apply based on the IF conditional?

Here's how I'm using this today: https://play.golang.org/p/WfnIAOzwZtC 

But, I've had instances where it seems the update has failed to apply, but returned cleanly and allowed the rest of my code to execute. I'm using this in a job locking mechanism and I've had the same job run twice on different workers.

PR #99, opened 2019-04-23
Naive attempt at implementing CAS support

Linuturk:
Per #98

### DIFF
--- a/queryx.go
+++ b/queryx.go
@@ -192,11 +192,28 @@ func (q *Queryx) Exec() error {
 	return q.Query.Exec()
 }
 
+// ExecCAS executes the query without returning any rows. Also returns
+// whether the query was applied.
+func (q *Queryx) ExecCAS() (bool, error) {
+	if q.err != nil {
+		return false, q.err
+	}
+	dest := map[string]interface{}{}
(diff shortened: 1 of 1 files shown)


Linuturk:
I've applied these changes to my internal project and I am testing it now. I'm not sure of the ramifications of these changes and am looking for feedback.

dahankzter:
This looks good but lacking a bit I think. Perhaps we should also implement the other CAS operations as provided by the driver? Also if at all possible we should also have tests for these operations. I assume you are running Cassandra of some recent version? Scylla is in the process of implementing LWT so that caveat we should perhaps list somewhere in the docs as well.

mmatczuk:
I do not think that solves it, as [Queryx embeds Query](https://github.com/scylladb/gocqlx/blob/master/queryx.go#L85) you can call `MapScanCAS` on Queryx directly. What is needed is CAS versions of BindStruct, Get etc.

Linuturk:
I'd agree with @mmatczuk based on my testing yesterday. With my changes every CAS query comes back with `applied` equal to `false`.

mmatczuk:
Fixed in #147

PR #102, opened 2019-05-13
Tuple support

dahankzter:
This adds tuple support to `qb`. It adds a number of versions of the comparators and builder primitifes that emit placeholders.

### DIFF
--- a/qb/delete_test.go
+++ b/qb/delete_test.go
@@ -43,6 +43,24 @@ func TestDeleteBuilder(t *testing.T) {
 			S: "DELETE FROM cycling.cyclist_name WHERE id=? AND firstname>? ",
 			N: []string{"expr", "firstname"},
 		},
+		// Add a tuple column
+		{
+			B: Delete("cycling.cyclist_name").Where(EqTuple("id", 2)).Columns("stars"),
+			S: "DELETE stars FROM cycling.cyclist_name WHERE id=(?,?) ",
+			N: []string{"id"},
+		},
+		// Add WHERE for tuple column
+		{
+			B: Delete("cycling.cyclist_name").Where(w, GtTuple("firstname", 2)),
+			S: "DELETE FROM cycling.cyclist_name WHERE id=? AND firstname>(?,?) ",
+			N: []string{"expr", "firstname"},
+		},
+		// Add WHERE for all tuple columns
+		{
+			B: Delete("cycling.cyclist_name").Where(EqTuple("id", 2), GtTuple("firstname", 2)),
+			S: "DELETE FROM cycling.cyclist_name WHERE id=(?,?) AND firstname>(?,?) ",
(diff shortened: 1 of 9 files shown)


qb/delete_test.go
```diff
@@ -13,12 +13,14 @@ import (
...
 func TestDeleteBuilder(t *testing.T) {
 	w := EqNamed("id", "expr")
+	wt := EqTuple("id", 2)
 
 	table := []struct {
 		B *DeleteBuilder
 		N []string
 		S string
 	}{
+
```
mmatczuk: new line not needed

qb/delete_test.go
```diff
@@ -13,12 +13,14 @@ import (
 
 func TestDeleteBuilder(t *testing.T) {
 	w := EqNamed("id", "expr")
+	wt := EqTuple("id", 2)
```
mmatczuk: Maybe inline?

qb/delete_test.go
```diff
@@ -43,6 +45,30 @@ func TestDeleteBuilder(t *testing.T) {
 			S: "DELETE FROM cycling.cyclist_name WHERE id=? AND firstname>? ",
 			N: []string{"expr", "firstname"},
 		},
+		// Add tuple column
```
mmatczuk: What is the point in cloning all the tests but with tuple? Maybe one or two tests could be just fine?

qb/select_test.go
```diff
@@ -12,6 +12,7 @@ import (
 
 func TestSelectBuilder(t *testing.T) {
 	w := EqNamed("id", "expr")
+	wt := EqTuple("id", 2)
```
mmatczuk: Maybe inline?

qb/insert_test.go
```diff
@@ -70,6 +70,18 @@ func TestInsertBuilder(t *testing.T) {
 			S: "INSERT INTO cycling.cyclist_name (id,user_uuid,firstname) VALUES (?,?,?) USING TIMESTAMP ? ",
 			N: []string{"id", "user_uuid", "firstname", "ts"},
 		},
+		// Add TupleColumn
```
mmatczuk: Maybe `Add a tuple column`

qb/insert_test.go
```diff
@@ -70,6 +70,18 @@ func TestInsertBuilder(t *testing.T) {
...
+		{
+			B: Insert("cycling.cyclist_name").TupleColumn("id", 2),
+			S: "INSERT INTO cycling.cyclist_name (id) VALUES ((?,?)) ",
+			N: []string{"id"},
+		},
+		// Add TupleColumn
```
mmatczuk: See above

qb/delete_test.go
dahankzter: I was unsure of the code at that point. Perhaps you are right.

PR #105, opened 2019-05-30
iterx: paging iteration working

dahankzter:
We used to rely upon NumRows to determine if there are
new pages available. This is not correct and the needed
data to do this is not exposed by the gocql driver.

We simply remove these checks and let the driver decide.

### DIFF
--- a/.travis_start_db.sh
+++ b/.travis_start_db.sh
@@ -4,7 +4,7 @@ set -e
 case ${DB} in
 
 scylla)
-    sudo curl -o /etc/apt/sources.list.d/scylla.list -L http://repositories.scylladb.com/scylla/repo/20fc70b18261bf832cf8e0733a27979c/ubuntu/scylladb-2.1-trusty.list
+    sudo curl -o /etc/apt/sources.list.d/scylla.list -L http://repositories.scylladb.com/scylla/repo/20fc70b18261bf832cf8e0733a27979c/ubuntu/scylladb-3.0-trusty.list
     sudo apt-get -qq update
     sudo apt-get install -y --allow-unauthenticated scylla-server
     sudo /usr/bin/scylla --options-file /etc/scylla/scylla.yaml ${SCYLLA_OPTS} &> /tmp/scylla.log &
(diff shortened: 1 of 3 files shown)


iterx.go
```diff
@@ -180,7 +170,13 @@ func (iter *Iterx) scanAll(dest interface{}, structOnly bool) bool {
 
 		// allocate memory for the page data
 		if !alloc {
-			v = reflect.MakeSlice(slice, 0, iter.Iter.NumRows())
+			sz := iter.NumRows()
+			if sz == 0 {
+				// Taken out of the air.
+				// Should it be the default page size?
+				sz = 100
```
mmatczuk: Note that users may change that https://godoc.org/github.com/gocql/gocql#Query.PageSize so it should be a default unless changed.

mmatczuk: Not sure exactly why NumRows is wrong, maybe it's an gocql error?

iterx_test.go
```diff
@@ -327,3 +328,46 @@ func TestNotFound(t *testing.T) {
...
+		defer it.Close()
+		var cnt int
+		for {
+			p := &Paging{}
+			if !it.StructScan(p) {
+				break
+			}
+			cnt++
+		}
+		if cnt != 100 {
```
mmatczuk: If default page size is 5k do we ever hit the bug without the patch here.

.travis_start_db.sh
```diff
@@ -4,7 +4,7 @@ set -e
 case ${DB} in
 
 scylla)
-    sudo curl -o /etc/apt/sources.list.d/scylla.list -L http://repositories.scylladb.com/scylla/repo/20fc70b18261bf832cf8e0733a27979c/ubuntu/scylladb-2.1-trusty.list
+    sudo curl -o /etc/apt/sources.list.d/scylla.list -L http://repositories.scylladb.com/scylla/repo/20fc70b18261bf832cf8e0733a27979c/ubuntu/scylladb-3.0-trusty.list
```
mmatczuk: We can piggyback that it could go as a separate PR.

iterx_test.go
dahankzter: I can't say for sure but it seems to vary with the workload and data set size.

.travis_start_db.sh
dahankzter: Ok. Putting up another PR.

PR #113, opened 2019-09-16
Add support for BYPASS CACHE

martin-sucha:
This commit adds support for BYPASS CACHE extension that is present in
ScyllaDB since 3.1 / 2019.1.1.

https://docs.scylladb.com/getting-started/dml/#bypass-cache

### DIFF
--- a/qb/select.go
+++ b/qb/select.go
@@ -40,6 +40,7 @@ type SelectBuilder struct {
 	limit             uint
 	limitPerPartition uint
 	allowFiltering    bool
+	bypassCache       bool
 	json              bool
 }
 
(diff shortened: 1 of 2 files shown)


rkuska:
LGTM

dahankzter:
It does look good. Any drawbacks other than that Cassandra users need to avoid it? @mmatczuk what do you think? Do we promise anything on optional methods like this?

dahankzter:
Perhaps you can add that this is a scylla specific ting and even add the link https://docs.scylladb.com/getting-started/dml/#bypass-cache in the doc @martin-sucha ?

qb/select.go
```diff
@@ -180,6 +185,12 @@ func (b *SelectBuilder) AllowFiltering() *SelectBuilder {
 	return b
 }
 
+// BypassCache sets a BYPASS CACHE clause on the query.
```
mmatczuk: In my view we should document that this is a Scylla only feature.

mmatczuk:
LGTM see the inline comment.

martin-sucha:
I've updated the docs as requested.

mmatczuk:
Thanks @martin-sucha

PR #115, opened 2019-10-27
Improve assign in qb package

meox:
Changed some core functions to avoid extra allocation performed during assigment, for instance:
Instead of recreate the slice in this way:
```
// Columns adds result columns to the query.
func (b *SelectBuilder) Columns(columns ...string) *SelectBuilder {
	b.columns = append(b.columns, columns...)
	return b
}
```

we can forwad it:

```
// Columns adds result columns to the query.
func (b *SelectBuilder) Columns(columns ...string) *SelectBuilder {
	if len(b.columns) == 0 {
		b.columns = columns
	} else {
		b.columns = append(b.columns, columns...)
	}
	return b
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@
 
 # Go vendor dir
 vendor
+
+# GoLand IDE
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,5 @@
 
 # Go vendor dir
 vendor
-
 # GoLand IDE
 .idea/
--- a/qb/select.go
+++ b/qb/select.go
@@ -132,7 +132,11 @@ func (b *SelectBuilder) Json() *SelectBuilder {
 
 // Columns adds result columns to the query.
 func (b *SelectBuilder) Columns(columns ...string) *SelectBuilder {
-	b.columns = append(b.columns, columns...)
+	if len(b.columns) == 0 {
+		b.columns = columns
+	} else {
+		b.columns = append(b.columns, columns...)
+	}
 	return b
 }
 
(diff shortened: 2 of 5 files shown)


go.sum
```diff
@@ -25,6 +25,8 @@ github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZb
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/scylladb/go-reflectx v1.0.1 h1:b917wZM7189pZdlND9PbIJ6NQxfDPfBvUaQ7cjj1iZQ=
 github.com/scylladb/go-reflectx v1.0.1/go.mod h1:rWnOfDIRWBGN0miMLIcoPt/Dhi2doCMZqwMCJ3KupFc=
+github.com/scylladb/gocqlx v1.3.1 h1:NTiKaSW1RzDxHQIyPE/KubOJCKRG5xXMUG8FKVKR/j0=
```
mmatczuk: ?

go.mod
```diff
@@ -12,3 +12,5 @@ require (
 	github.com/stretchr/testify v1.3.0 // indirect
 	gopkg.in/inf.v0 v0.9.1
 )
+
```
mmatczuk: That's good can we have that in a separate commit.

meox: reverted change on go.mod & go.sum

go.sum
meox: removed

mmatczuk:
* The `Set` naming is much better
* For allocation why not just assign when nil?
* I think this should be added to all the builders
* Tests needs to be added

meox:
>     * The `Set` naming is much better
> 
so like now?

>     * For allocation why not just assign when nil?
> 
I'm not following... the purpose of this PR is to avoid the usage of append function

>     * I think this should be added to all the builders
> 
do you have a list?

>     * Tests needs to be added

mmatczuk:
@meox in my view this change only pollutes the API. One change that we could make all over the `qb` builders is the following:

Instead of just `x.x = append(x.x, y...)` we can do

```go
if len(x.x) == 0 {
    x.x = y
} else {
    x.x = append(x.x, y...)
}
```

Again this can be applied consistently to all the `qb` code whenever we append slices.

meox:
if we do this we break the API

.gitignore
```diff
@@ -10,5 +10,12 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+
```
mmatczuk: That does not belong to this PR.

meox:
```
goos: linux
goarch: amd64
pkg: github.com/scylladb/gocqlx/qb
BenchmarkBatchBuilder-32                 2027047               594 ns/op
BenchmarkCmp-32                          1978915               607 ns/op
BenchmarkDeleteBuilder-32                1627681               733 ns/op
BenchmarkInsertBuilder-32                1000000              1146 ns/op
BenchmarkSelectBuilder-32                1956966               614 ns/op
BenchmarkSelectBuildAssign-32            2176251               554 ns/op
BenchmarkUpdateBuilder-32                 776941              1545 ns/op
PASS
```

mmatczuk:
So what is the difference?

meox:
> So what is the difference?

10% faster with few columns, we have many columns so I expect more in a real scenario

.gitignore
```diff
@@ -12,3 +12,6 @@
 
 # Go vendor dir
 vendor
+
```
mmatczuk: Please remove that.

meox: the empty line?

meox: done

PR #117, opened 2019-11-06
Test with Go 1.12 and 1.13

martin-sucha:
Support last two versions of Go, similar to gocql. We also need to
remove the flag.Parse() call because of the change in Go 1.13:

> Testing flags are now registered in the new Init function, which is
> invoked by the generated main function for the test. As a result,
> testing flags are now only registered when running a test binary, and
> packages that call flag.Parse during package initialization may cause
> tests to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 sudo: true
 language: go
 go:
-  - 1.11.x
+  - stable
+  - oldstable
 
 env:
   global:
--- a/Makefile
+++ b/Makefile
@@ -40,4 +40,4 @@ get-deps:
 .PHONY: get-tools
 get-tools:
 	@echo "==> Installing tools at $(GOBIN)..."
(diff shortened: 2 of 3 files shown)


dahankzter:
@martin-sucha I think that Megacheck needs a bump too?

.travis.yml
```diff
@@ -1,7 +1,8 @@
 sudo: true
 language: go
 go:
-  - 1.11.x
+  - 1.12.x
+  - 1.13.x
```
mmatczuk: To avoid that in the future I think we could just use `tip`.

martin-sucha: That's a good idea. In Travis CI docs they say https://github.com/travis-ci/gimme is used to get the Go version, so we could use `stable` and `oldstable` instead of specific versions.

mmatczuk:
Fixed with #122

PR #118, opened 2019-11-06
Enhance support for scanning UDTs and entire rows into same struct type

martin-sucha:


iterx.go
```diff
@@ -108,13 +103,13 @@ func (iter *Iterx) scanAny(dest interface{}, structOnly bool) bool {
...
 func (iter *Iterx) Select(dest interface{}) error {
-	iter.scanAll(dest, false)
+	iter.scanAll(dest)
 	iter.Close()
 
 	return iter.err
 }
 
-func (iter *Iterx) scanAll(dest interface{}, structOnly bool) bool {
+func (iter *Iterx) scanAll(dest interface{}) bool {
```
dahankzter: Here the boolean argument is removed but in the other commit it's brought back?

iterx.go
```diff
@@ -103,13 +103,24 @@ func (iter *Iterx) scanAny(dest interface{}) bool {
...
+//
+// If no rows were selected, ErrNotFound is NOT returned.
+func (iter *Iterx) StructSelect(dest interface{}) error {
+	iter.scanAll(dest, true)
+	iter.Close()
+
+	return iter.err
+}
+
+func (iter *Iterx) scanAll(dest interface{}, structScan bool) bool {
```
dahankzter: Here the boolean is brought back?

iterx.go
```diff
@@ -132,10 +143,17 @@ func (iter *Iterx) scanAll(dest interface{}) bool {
 	base := reflectx.Deref(slice.Elem())
 	scannable := isScannable(base)
 
-	// if it's a base type make sure it only has 1 column;  if not return an error
-	if scannable && len(iter.Columns()) > 1 {
-		iter.err = fmt.Errorf("non-struct dest type %s with >1 columns (%d)", base.Kind(), len(iter.Columns()))
-		return false
+	if structScan {
```
dahankzter: If `structScan` is true there is no need to check if it's scannable right? Is there a chance for an early return when the user wants to scan into a struct?

gocqlx.go
```diff
@@ -13,23 +13,10 @@ import (
...
-		return fmt.Errorf("structscan expects a struct dest but the provided struct type %s implements unmarshaler", t.Name())
-	}
-	return fmt.Errorf("expected a struct, but struct %s has no exported fields", t.Name())
-}
-
 // reflect helpers
 
-var _unmarshallerInterface = reflect.TypeOf((*gocql.Unmarshaler)(nil)).Elem()
+var unmarshallerInterface = reflect.TypeOf((*gocql.Unmarshaler)(nil)).Elem()
+var udtUnmarshallerInterface = reflect.TypeOf((*gocql.Unmarshaler)(nil)).Elem()
```
mmatczuk: Is that used, it's `gocql.Unmarshaler`? in general var's should be grouped.

mmatczuk:
@martin-sucha I'd much rather like to add that as a tweak in Iterx similar to https://github.com/scylladb/gocqlx/blob/master/iterx.go#L54. This change as it is makes API less readable.

mmatczuk:
Any updates on that?

martin-sucha:
> Any updates on that?

I've been busy this week, sorry. I'll take a look at this next week.

martin-sucha:
> > Any updates on that?
> 
> I've been busy this week, sorry. I'll take a look at this next week.

Took longer than expected, but finally updated now.

gocqlx.go
```diff
@@ -13,23 +13,10 @@ import (
...
-	if isScanner {
-		return fmt.Errorf("structscan expects a struct dest but the provided struct type %s implements unmarshaler", t.Name())
-	}
-	return fmt.Errorf("expected a struct, but struct %s has no exported fields", t.Name())
-}
-
 // reflect helpers
 
-var _unmarshallerInterface = reflect.TypeOf((*gocql.Unmarshaler)(nil)).Elem()
+var unmarshallerInterface = reflect.TypeOf((*gocql.Unmarshaler)(nil)).Elem()
```
mmatczuk: Group maybe.

gocqlx.go
```diff
@@ -42,10 +29,11 @@ func baseType(t reflect.Type, expected reflect.Kind) (reflect.Type, error) {
 // isScannable takes the reflect.Type and the actual dest value and returns
 // whether or not it's Scannable. Something is scannable if:
 //   * it is not a struct
-//   * it implements gocql.Unmarshaler
+//   * it implements gocql.Unmarshaler or gocql.UDTUnmarshaler
 //   * it has no exported fields
 func isScannable(t reflect.Type) bool {
-	if reflect.PtrTo(t).Implements(_unmarshallerInterface) {
+	ptr := reflect.PtrTo(t)
```
mmatczuk: How about `if ptr := ...; ptr.Implements(...`?

gocqlx.go
martin-sucha: Updated.

gocqlx.go
martin-sucha: Updated.

iterx.go
```diff
@@ -56,20 +58,34 @@ func (iter *Iterx) Unsafe() *Iterx {
...
 }
 
-// Get scans first row into a destination and closes the iterator. If the
-// destination type is a struct pointer, then StructScan will be used.
-// If the destination is some other type, then the row must only have one column
-// which can scan into that type.
+// Struct forces the iterator to treat a single-argument struct as non-scannable.
+// This is is useful if you need to scan a row into a struct that also implements gocql.Unmarshaler
+// or gocql.UDTUnmarshaler.
+func (iter *Iterx) Struct() *Iterx {
```
mmatczuk: It's not very consistent, I think we should name it like `DisableUDT` both field and func.

iterx.go
```diff
@@ -56,20 +58,34 @@ func (iter *Iterx) Unsafe() *Iterx {
...
+//
+// If the destination type is scannable (non-struct, gocql.Unmarshaler, gocql.Marshaler), the row must have only
+// one column which can scan into that type.
+// If the destination type is non-scannable struct pointer or Struct() was used on the iterator, then
+// StructScan will be used.
 //
 // If no rows were selected, ErrNotFound is returned.
 func (iter *Iterx) Get(dest interface{}) error {
-	iter.scanAny(dest, false)
+	if iter.forceStruct {
```
mmatczuk: This check should be moved to scanAny?

mmatczuk:
Hi @martin-sucha sorry it took so long and thanks for your patience.

It's a good idea and a valid change - thank you!

I have rebaseds and refactored your proposal please take a look at https://github.com/scylladb/gocqlx/pull/143.

I'm closing this for now. Work will continue in 143.

Issue #130, opened 2020-01-13
Guidance on missing destination name error


annismckenzie:
I just ran into this with one of our production services where we did a migration beforehand and all of a sudden all requests were failing on the running services with the error

> missing destination name "customer_id" in *model.Company

I didn't know that this would happen. After investigating a little bit I could've used `.Unsafe()` on `Iterx` or explicitly list the columns to select using `Columns()`.

My question would be what the general guidance on this is? Why is the method called `Unsafe`? We transform our models into protobuf objects so there's no inherent security issue at play.

I'm also wary of having to go through dozens of services and checking each query so that it either has `Unsafe()` called or explicitly calls `Columns()`. I didn't know this would happen even though it was the first time it actually bit us.

Would there be a way to add an option to enable `unsafe` for the whole app? How do you handle migrations in your services?

mmatczuk:
We use table if possible https://github.com/scylladb/gocqlx/blob/master/table/example_test.go this way if removing a column we remove it from metadata and it's done.

Unsafe is to inform you that you load more data than you can consume.

annismckenzie:
> We use table if possible https://github.com/scylladb/gocqlx/blob/master/table/example_test.go this way if removing a column we remove it from metadata and it's done.

Sounds fair enough for your use case. This feature should be mentioned somewhere.

> Unsafe is to inform you that you load more data than you can consume.

Hmmm, for single primary key lookups? That doesn't sound remotely unsafe and IMHO shouldn't blow up a select. There could be various reasons why one doesn't mention a specific column, be it intentional or unintentional. In our case it's usually to prepare an update of a schema. When I started using gocqlx the table package didn't even exist. Sigh, okay.

mmatczuk:
@annismckenzie you can try sending a PR with a global var `DefaultUnsafe` if that helps in porting your code.

annismckenzie:
That is great news! I'll get on that right away (I really want to kill our fork, it's driving me nuts 😅).

Re global var: any preference on making it an atomic?

mmatczuk:
Just a simple var

annismckenzie:
Thanks for accepting the PR and making my life easier! 🎉

PR #131, opened 2020-01-15
Add DefaultUnsafe global

annismckenzie:
As discussed in #130 here's a simple option to enable Iterx's `Unsafe()` behavior globally.

### DIFF
--- a/iterx_test.go
+++ b/iterx_test.go
@@ -276,6 +276,22 @@ func TestUnsafe(t *testing.T) {
 			t.Fatal("select failed")
 		}
 	})
+
+	t.Run("DefaultUnsafe select", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


iterx_test.go
```diff
@@ -276,6 +276,22 @@ func TestUnsafe(t *testing.T) {
 			t.Fatal("select failed")
 		}
 	})
+
+	t.Run("DefaultUnsafe select", func(t *testing.T) {
+		gocqlx.DefaultUnsafe = true
+		defer func() { gocqlx.DefaultUnsafe = true }()
```
annismckenzie: This should set it to false. 🤦‍♂

annismckenzie: done

annismckenzie:
Thank you!

PR #133, opened 2020-01-27
Await schema agreement during migrations

annismckenzie:
This uses `(*gocql.Session).AwaitSchemaAgreement` before applying each CQL statement in a migration (see https://github.com/gocql/gocql/issues/1301).

It's marked as a draft for discussion as well as because it's still waiting on https://github.com/gocql/gocql/pull/1395 to be merged and is currently using our fork of gocql.

### DIFF
--- a/migrate/migrate.go
+++ b/migrate/migrate.go
@@ -22,6 +22,26 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// DefaultAwaitSchemaAgreement controls whether checking for cluster schema agreement
+// is disabled or if it is checked before each file or statement is applied.
+// The default is not checking before each file or statement but only once after every
+// migration has been run.
+var DefaultAwaitSchemaAgreement = AwaitSchemaAgreementDisabled
+
+type awaitSchemaAgreement int
+
+// Options for checking schema agreement.
+const (
+	AwaitSchemaAgreementDisabled awaitSchemaAgreement = iota
+	AwaitSchemaAgreementBeforeEachFile
+	AwaitSchemaAgreementBeforeEachStatement
+)
+
+// ShouldAwait decides whether to await schema agreement for the configured DefaultAwaitSchemaAgreement option above.
+func (as awaitSchemaAgreement) ShouldAwait(stage awaitSchemaAgreement) bool {
+	return as == stage
+}
+
 const (
 	infoSchema = `CREATE TABLE IF NOT EXISTS gocqlx_migrate (
 	name text,
@@ -117,6 +137,10 @@ func Migrate(ctx context.Context, session *gocql.Session, dir string) error {
 		}
 	}
 
+	if err = session.AwaitSchemaAgreement(ctx); err != nil {
+		return fmt.Errorf("awaiting schema agreement failed: %s", err)
+	}
+
 	return nil
 }
 
@@ -149,6 +173,12 @@ func applyMigration(ctx context.Context, session *gocql.Session, path string, do
 	iq := gocqlx.Query(session.Query(stmt).WithContext(ctx), names)
 	defer iq.Release()
 
+	if DefaultAwaitSchemaAgreement.ShouldAwait(AwaitSchemaAgreementBeforeEachFile) {
+		if err = session.AwaitSchemaAgreement(ctx); err != nil {
+			return fmt.Errorf("awaiting schema agreement failed: %s", err)
+		}
+	}
+
 	i := 0
 	r := bytes.NewBuffer(b)
 	for {
@@ -173,6 +203,12 @@ func applyMigration(ctx context.Context, session *gocql.Session, path string, do
 		if Callback != nil && i == 1 {
 			if err := Callback(ctx, session, BeforeMigration, info.Name); err != nil {
 				return fmt.Errorf("before migration callback failed: %s", err)
+			}
+		}
+
+		if DefaultAwaitSchemaAgreement.ShouldAwait(AwaitSchemaAgreementBeforeEachStatement) {
+			if err = session.AwaitSchemaAgreement(ctx); err != nil {
+				return fmt.Errorf("awaiting schema agreement failed: %s", err)
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


migrate/migrate.go
```diff
@@ -22,6 +22,10 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// AwaitSchemaAgreementBeforeApplyingStatement controls whether to await
+// schema agreement before applying each migration statement.
+var AwaitSchemaAgreementBeforeApplyingStatement = true
```
annismckenzie: Does this even have value? I'd like to get rid of it, don't know what I was thinking.

mmatczuk:
This needs to be configurable with the following options

* Disabled
* Enabled -> After each file

There is no point in applying it after each statement (sent to the same host) or before anything.

annismckenzie:
Hmmm, but I don't want to apply even one statement if the cluster is in a bad state (nodes unavailable). Also, I'm usually putting both dropping a materialized view and recreating it into the same migration and these statements should only be applied one after another with the schema agreed cluster-wide. I already have this version of gocqlx in production right now. The same host doesn't really matter if all nodes agree on the schema or am I missing something?

With the current variable at the top I only need to add it to the final call after all the migrations have successfully been applied. WDYT?

mmatczuk:
The final call should be mandatory.

We can have the enum extended to 
* Disabled
* Enabled -> After each file
* Enabled -> After each stmt

annismckenzie:
If I get you to agree to change it to this:
- Disabled
- Enabled -> _Before_ each file
- Enabled -> _Before_ each stmt

Then  I would agree. I'm reiterating: I don't want to apply a single statement if the cluster is an unhealthy state or doesn't have schema agreement. With your proposal that wouldn't be possible or require me to add another one before starting the migrations (something I don't have to do if I change both modes to check before instead of after).

mmatczuk:
Optional before (file or statement) plus mandatory after full migration is perfect.

annismckenzie:
Gave it a go with an enum. I hope you like it. :)

migrate/migrate.go
```diff
@@ -117,6 +138,10 @@ func Migrate(ctx context.Context, session *gocql.Session, dir string) error {
 		}
 	}
 
+	if err = AwaitSchemaAgreementOption.Check(ctx, session, awaitSchemaAgreementAfterMigrations); err != nil {
```
annismckenzie: I added the private `awaitSchemaAgreementAfterMigrations` option to unify the execution of AwaitSchemaAgreement on the gocql session (see line 40 for the check). I can also drop that and instead just put the three lines here again.

migrate/migrate.go
```diff
@@ -22,6 +22,27 @@ import (
 	"github.com/scylladb/gocqlx/qb"
 )
 
+// AwaitSchemaAgreementOption controls whether checking for cluster schema agreement
+// is disabled or if it is checked before each file or statement is applied.
+var AwaitSchemaAgreementOption awaitSchemaAgreement
```
mmatczuk: `var DefaultAwaitSchemaAgreement = AwaitSchemaAgreementDisabled`

migrate/migrate.go
```diff
@@ -22,6 +22,27 @@ import (
...
+
+// Options for checking schema agreement.
+const (
+	AwaitSchemaAgreementDisabled awaitSchemaAgreement = iota
+	AwaitSchemaAgreementBeforeEachFile
+	AwaitSchemaAgreementBeforeEachStatement
+	awaitSchemaAgreementAfterMigrations
+)
+
+func (as awaitSchemaAgreement) Check(ctx context.Context, session *gocql.Session, stage awaitSchemaAgreement) error {
```
mmatczuk: Docs

mmatczuk: In my view it would be better to separate the fileterig (SchouldAwait) logic from the `session.AwaitSchemaAgreement` call.

migrate/migrate.go
mmatczuk: L40 is ok with regards to that.

migrate/migrate.go
annismckenzie: Yeah, sorry about that – done.

migrate/migrate.go
annismckenzie: done

migrate/migrate.go
annismckenzie: Oh damn, didn't see your second comment there.

annismckenzie: Not sure what you mean I should change. Can you elaborate? 😅

mmatczuk: Take out `session.AwaitSchemaAgreement(ctx)` from this function and rename to `SchouldAwait`.

migrate/migrate.go
```diff
@@ -22,6 +22,30 @@ import (
...
+var DefaultAwaitSchemaAgreement = AwaitSchemaAgreementDisabled
+
+type awaitSchemaAgreement int
+
+// Options for checking schema agreement.
+const (
+	AwaitSchemaAgreementDisabled awaitSchemaAgreement = iota
+	AwaitSchemaAgreementBeforeEachFile
+	AwaitSchemaAgreementBeforeEachStatement
+	awaitSchemaAgreementAfterMigrations
```
mmatczuk: There is no need to check it.

migrate/migrate.go
annismckenzie: All done. Take another look? I liked the previous version better because it saved having to use 2 nested `if`s but it's your call. :)

migrate/migrate.go
annismckenzie: yup, that's gone now

annismckenzie:
Running this one last time in our infra inside a GitLab CI job with `DefaultAwaitSchemaAgreement` set to `AwaitSchemaAgreementBeforeEachStatement` now. 🤞

annismckenzie:
Works as advertised. 🎉 MIgrations take a bit longer than before (which is expected) but I'm now confident that I won't run into any problems with migrations ever again (apart from them failing because of node issues).

mmatczuk:
Thanks.

PR #137, opened 2020-03-13
gocqlx/table/Table: support timestamp at the granularity of batch statements

GingerMoon:
I'd like to add some public functions to expose CRUD query builders,  which is similar as
func (t *Table) SelectBuilder(columns ...string) *qb.SelectBuilder {
to enable timestamp at the granularity of batch statements.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -105,10 +105,20 @@ func (t *Table) Insert() (stmt string, names []string) {
 
 // Update returns update by primary key statement.
 func (t *Table) Update(columns ...string) (stmt string, names []string) {
-	return qb.Update(t.metadata.Name).Set(columns...).Where(t.primaryKeyCmp...).ToCql()
+	return t.UpdateBuilder(columns...).ToCql()
+}
+
+// UpdateBuilder returns a builder initialised to update by primary key statement.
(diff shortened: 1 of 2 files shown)


GingerMoon:
Hi @mmatczuk do you know who I should contact for reviewing this PR?

mmatczuk:
Hi @GingerMoon thanks for pinging me.

table/table.go
```diff
@@ -28,7 +28,7 @@ type Table struct {
 
 	get    cql
 	sel    cql
-	insert cql
```
mmatczuk: I'd rather we did not change that.

table/table.go
```diff
@@ -100,15 +97,30 @@ func (t *Table) SelectBuilder(columns ...string) *qb.SelectBuilder {
 
 // Insert returns insert all columns statement.
 func (t *Table) Insert() (stmt string, names []string) {
-	return t.insert.stmt, t.insert.names
+	return t.InsertBuilder().ToCql()
+}
+
+// InsertBuilder returns a builder initialised to insert all columns statement.
+func (t *Table) InsertBuilder() *qb.InsertBuilder {
```
mmatczuk: The builders are mutable you cannot return a pointer like that.

It should be implemented as UpdateBuilder and DeleteBuilder.

PR #147, opened 2020-04-20
queryx: Added CAS exec functions (#98)

### LINKED ISSUE #98: Support for CAS Operation Insight (LWT)

zimnx:
Added ExecCAS and ExecCASRelease functions suitable for INSERT ... IF
NOT EXISTS and UPDATE's containing IF statement.
Functions returns information wheter query was applied or not.

Fixes #98

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -24,6 +24,7 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
     first_name text,
     last_name text,
     email list<text>,
+    salary int,
     PRIMARY KEY(first_name, last_name)
 )`
 
@@ -38,12 +39,14 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
 		FirstName string
 		LastName  string
 		Email     []string
+		Salary    int
 	}
 
 	p := Person{
 		"Patricia",
 		"Citizen",
 		[]string{"patricia.citzen@gocqlx_test.com"},
+		500,
 	}
 
 	// Insert, bind data from struct.
@@ -117,11 +120,13 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
 				"Igy",
 				"Citizen",
 				[]string{"igy.citzen@gocqlx_test.com"},
+				500,
 			},
 			B: Person{
 				"Ian",
 				"Citizen",
 				[]string{"ian.citzen@gocqlx_test.com"},
+				500,
 			},
 		}
 		q := session.Query(stmt, names).BindStruct(&batch)
@@ -171,6 +176,7 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
 			"Ian",
 			"Citizen",
 			[]string{"ian.citzen@gocqlx_test.com"},
+			500,
 		}
 
 		stmt, names := qb.Select("gocqlx_test.person").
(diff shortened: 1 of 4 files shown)


example_test.go
```diff
@@ -201,11 +207,54 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
...
+			[]string{"stephen.johns@gocqlx_test.com"},
+			500,
+		}
+
+		stmt, names := qb.Insert("gocqlx_test.person").
+			Columns("first_name", "last_name", "email", "salary").
+			Unique().
+			ToCql()
+
+		applied, err := session.Query(stmt, names).BindStruct(p).ExecCASRelease()
```
mmatczuk, line 234: Exec?

example_test.go
```diff
@@ -201,11 +207,54 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		t.Log(applied)
+
+		stmt, names = qb.Update("gocqlx_test.person").
+			SetNamed("salary", "new_salary").
+			If(qb.Lt("salary")).
```
mmatczuk, line 243: If goes after where

example_test.go
```diff
@@ -201,11 +207,54 @@ CREATE TABLE IF NOT EXISTS gocqlx_test.person (
...
+			Where(qb.Eq("first_name"), qb.Eq("last_name")).
+			ToCql()
+		q := session.Query(stmt, names).BindMap(qb.M{
```
mmatczuk, line 246: I'd suggest to use BindStructMap here

iterx.go
```diff
@@ -257,27 +258,39 @@ func (iter *Iterx) structScan(value reflect.Value) bool {
 		panic("value must be a pointer")
 	}
 
+	casResponse := false
 	if iter.fields == nil {
 		columns := columnNames(iter.Iter.Columns())
-		iter.fields = iter.Mapper.TraversalsByName(value.Type(), columns)
+		if len(columns) > 0 && columns[0] == "[applied]" {
```
mmatczuk, line 264: "[applied]" should be a const

iterx.go
```diff
@@ -257,27 +258,39 @@ func (iter *Iterx) structScan(value reflect.Value) bool {
...
+		iter.fields = iter.Mapper.TraversalsByName(value.Type(), columns)
 		// if we are not unsafe and are missing fields, return an error
 		if !iter.unsafe {
 			if f, err := missingFields(iter.fields); err != nil {
 				iter.err = fmt.Errorf("missing destination name %q in %s", columns[f], reflect.Indirect(value).Type())
 				return false
 			}
 		}
-		iter.values = make([]interface{}, len(columns))
+		iter.values = make([]interface{}, len(columns), cap(columns))
```
mmatczuk, line 278: Why?

iterx.go
```diff
@@ -257,27 +258,39 @@ func (iter *Iterx) structScan(value reflect.Value) bool {
 		panic("value must be a pointer")
 	}
 
+	casResponse := false
```
mmatczuk, line 261: This only work for the 1st row!

It would be better to prepend `iter.fields` with nil to accomodate for applied and and init iter.values[0] in the if branch below.

mmatczuk: I'd rename that to just cas btw.

queryx.go
```diff
@@ -209,6 +209,23 @@ func (q *Queryx) ExecRelease() error {
 	return q.Exec()
 }
 
+// ExecCAS executes the Lightweight Transaction query, returns whether query was applied.
+// It's suitable for INSERT ... IF NOT EXISTS and UPDATE's having IF conditions.
```
mmatczuk, line 213: A link to Scylla docs on LWT would be more correct.

queryx.go
```diff
@@ -236,6 +253,29 @@ func (q *Queryx) GetRelease(dest interface{}) error {
 	return q.Get(dest)
 }
 
+// GetCAS executes a lightweight transaction (i.e. an UPDATE or INSERT
```
mmatczuk, line 256: Again add LWD docs link and let's not get into too many details.

queryx.go
```diff
@@ -236,6 +253,29 @@ func (q *Queryx) GetRelease(dest interface{}) error {
...
+// statement containing an IF clause). If the transaction fails because
+// the existing values did not match, the previous values will be stored
+// in dest object.
+//
+// As for INSERT .. IF NOT EXISTS, previous values will be returned as if
+// SELECT * FROM.
```
mmatczuk, line 262: with "SELECT * FROM ..." query

mmatczuk:
Thanks!

PR #150, opened 2020-04-21
queryx: Add test to check that all methods are wrapped

mmatczuk:
Signed-off-by: Michał Matczuk <michal@scylladb.com>

### DIFF
--- a/queryx_test.go
+++ b/queryx_test.go
@@ -5,8 +5,10 @@
 package gocqlx
 
 import (
+	"reflect"
 	"testing"
 
+	"github.com/gocql/gocql"
 	"github.com/google/go-cmp/cmp"
 )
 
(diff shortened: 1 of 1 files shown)


queryx_test.go
```diff
@@ -149,3 +151,25 @@ func TestQueryxBindMap(t *testing.T) {
...
+func TestQyeryxAllWrapped(t *testing.T) {
+	var (
+		gocqlQueryPtr = reflect.TypeOf((*gocql.Query)(nil))
+		queryxPtr     = reflect.TypeOf((*Queryx)(nil))
+	)
+
+	for i := 0; i < gocqlQueryPtr.NumMethod(); i++ {
+		m, ok := queryxPtr.MethodByName(gocqlQueryPtr.Method(i).Name)
+		if !ok {
+			t.Fatalf("Queryx missing method %s", gocqlQueryPtr.Method(i).Name)
```
zimnx, line 164: Having t.Errorf() here would end up with list of all missing methods. Fatal breaks test execution after first message.

PR #155, opened 2020-04-29
session: Added CQL schema dump feature

zimnx:


mmatczuk:
@zimnx let's move all this work to https://github.com/scylladb/gocql

Issue #156, opened 2020-05-12
table.go: support full table scan


GingerMoon:
I'd like to make a PR for add a full table scan function to table.go.
Basically it is inspred by [How to scan 475 million partitions 12x faster using efficient full table scan with Scylla 1.6](https://www.scylladb.com/2017/03/28/parallel-efficient-full-table-scan-scylla/)
Will you be interested in the PR?

mmatczuk:
It's implemented in https://github.com/scylladb/gocqlx/blob/master/example_test.go#L504 but sure we can port it as a gerenal pupruse function.

The main code corpus should live in gocqlx itself and should resolve parameters like `nodesInCluster` and `coresInNode` it may require patches to our [gocql fork](https://github.com/scylladb/gocql) first.

It should be then invoked by table module with table name.

@GingerMoon please send some sort of design.

GingerMoon:
I'd like to separate into several commits:
1. Simply port the the unit test. And make stuff like nodesInCluster as a configuration parameter.
2. Add stuff such as nodesInCluster into the scylladb and then change table.go accordingly.
3. Refine the commit#1. For example:
 There can be too many rows returns for some specific tokens.Then we need to retry with the "limit". 
 Add something like filter/handler callback to handle the returned row. The "producer" and "consumer" of the rows will work at the same time so that we don't need to keep all the returned rows.

mmatczuk:
> There can be too many rows returns for some specific tokens.Then we need to retry with the "limit".

In my view you want something like `func ForEach(table string, fn interface{})` GocqlX should ensure that the fn is sth like `func(v MyType) bool` and extract all the type info from it.

zwopir:
we have this implemented at JustWatch and we can opensource the code, if you are interested.
The library implements a scanner type with its central method `Scan` which is defined as
```go
// Scan performs a token range scan with the provided select builder
// The scanned items are passed to the ApplyFunc. If the ApplyFunc returns ErrAbort the scanning process
// is aborted. All other ApplyFunc errors are just logged
func (s *Scanner) Scan(
	ctx context.Context,
	selectBuilder *qb.SelectBuilder,
	partitionKeyColumns []string,
	applyFunc ApplyFunc,
) error
```
with `ApplyFunc` being defined as
```go
// ApplyFunc is a callback function that is called with each retrieved scan item
type ApplyFunc func(context.Context, map[string]interface{}) error
```
would you be interested?

zwopir:
[...]

PR #157, opened 2020-05-20
add a token scanner

zwopir:
this PR implements a token scanner. Follow up from https://github.com/scylladb/gocqlx/issues/156 Open for discussion


go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/scylladb/gocqlx/v2
 
 require (
-	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/gocql/gocql v0.0.0-20200131111108-92af2e088537
-	github.com/golang/snappy v0.0.1 // indirect
-	github.com/google/go-cmp v0.2.0
+	github.com/google/go-cmp v0.4.0
+	github.com/prometheus/client_golang v1.6.0
```
mmatczuk, line 6: This should be removed.

go.mod
```diff
@@ -1,11 +1,11 @@
...
 	github.com/scylladb/go-reflectx v1.0.1
+	github.com/scylladb/gocqlx v1.5.0
```
mmatczuk, line 8: Use gocqlx v2

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+	"context"
+	"math"
+	"math/rand"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/gocql/gocql"
+
+	"github.com/scylladb/gocqlx"
```
mmatczuk, line 13: Again v2

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"github.com/scylladb/gocqlx/qb"
+)
+
+const (
+	// parallelism = nodesInCluster * coresInNode * 3
+	defaultParallelism = 72
```
mmatczuk, line 21: ?

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+type tokenRange struct {
+	startRange int64
+	endRange   int64
+}
+
+// SessionBuilder creates a gocql.Session
```
mmatczuk, line 29: Single session should be used, there is no point in opening more TCP connections from one app.

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+type SessionBuilder interface {
+	CreateSession() (*gocql.Session, error)
+}
+
+// ApplyFunc is a callback function that is called with each retrieved scan item.
+// The scan is aborted if the ApplyFunc returns false
+type ApplyFunc func(context.Context, map[string]interface{}) bool
```
mmatczuk, line 36: I do not think that context should be passed here, implementers can use context of their own.
This implies usage of MapScan I'd like to see StructScan, therefore ApplyFunc should not exist and we should use interface{} like in cmp for instance.

```
// SortSlices returns a Transformer option that sorts all []V.
// The less function must be of the form "func(T, T) bool" which is used to
// sort any slice with element type V that is assignable to T.
//
// The less function must be:
//	• Deterministic: less(x, y) == less(x, y)
//	• Irreflexive: !less(x, x)
//	• Transitive: if !less(x, y) and !less(y, z), then !less(x, z)
//
// The less function does not have to be "total". That is, if !less(x, y) and
// !less(y, x) for two elements x and y, their relative order is maintained.
//
// SortSlices can be used in conjunction with EquateEmpty.
func SortSlices(lessFunc interface{}) cmp.Option {
	vf := reflect.ValueOf(lessFunc)
	if !function.IsType(vf.Type(), function.Less) || vf.IsNil() {
		panic(fmt.Sprintf("invalid less function: %T", lessFunc))
	}
	ss := sliceSorter{vf.Type().In(0), vf}
	return cmp.FilterValues(ss.filter, cmp.Transformer("cmpopts.SortSlices", ss.sort))
}
```

tokenscan/scanner.go
zwopir: leftover from our setup, sorry. Anything sensible we can use here? Otherwise I'd make it a mandatory parameter in the constructor

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+
+// Scanner is a scylla token range scanner
+type Scanner struct {
+	sessionBuilder SessionBuilder
+	tokenRanges    []*tokenRange
```
mmatczuk, line 41: There is no need to allocate memory for it, check the test producer consumer impl https://github.com/scylladb/gocqlx/blob/a72115182914eced6612440463a3ab40c727c5b4/example_test.go#L549

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+	parallelism    int
+	results        prometheus.Counter
+	errors         prometheus.Counter
+}
+
+// Opt is an option func
+type Opt func(*Scanner)
```
mmatczuk, line 48: Sadly scanner has no exported fields.

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+
+// WithParallelism configures the scan parallelism
+func WithParallelism(parallelism int) Opt {
```
mmatczuk, line 51: Let's just pass it as a function parameter.

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+type SessionBuilder interface {
+	CreateSession() (*gocql.Session, error)
+}
+
+// ApplyFunc is a callback function that is called with each retrieved scan item.
+// The scan is aborted if the ApplyFunc returns false
+type ApplyFunc func(context.Context, map[string]interface{}) bool
+
+// Scanner is a scylla token range scanner
+type Scanner struct {
```
mmatczuk, line 39: I doubt shis struct is need - Having just a function would be easier.

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+	s.errors.Collect(ch)
+}
+
+// Scan performs a token range scan with the provided select builder
+// The scanned items are passed to the ApplyFunc. If the ApplyFunc returns ErrAbort the scanning process
+// is aborted. All other ApplyFunc errors are just logged
+func (s *Scanner) Scan(
+	ctx context.Context,
+	selectBuilder *qb.SelectBuilder,
+	partitionKeyColumns []string,
```
mmatczuk, line 98: We should pass a struct to bind parameters from, we can pass the additional map to use BindStructMap this way one gets more flexibility on binding.

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+	applyFunc ApplyFunc,
```
mmatczuk, line 99: `apply interface{}` see my comment in L36

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+	for i := range s.tokenRanges {
+		rangesChannel <- s.tokenRanges[i]
+	}
+	close(rangesChannel)
+
+	// create a cancelable context
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	var sessionCreationWaitGroup sync.WaitGroup
+	errGroup := errgroup.Group{}
```
mmatczuk, line 112: var errGroup

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+) error {
+	// send all available token ranges to a channel
+	rangesChannel := make(chan *tokenRange, len(s.tokenRanges))
+	for i := range s.tokenRanges {
+		rangesChannel <- s.tokenRanges[i]
+	}
+	close(rangesChannel)
+
+	// create a cancelable context
+	ctx, cancel := context.WithCancel(ctx)
```
mmatczuk, line 109: errgroup handles context cancellation see https://godoc.org/golang.org/x/sync/errgroup#WithContext

tokenscan/scanner.go
zwopir: I implemented it as a struct to be able to implement the `prometheus.Collector` interface on it. If we don't want this (which I can understand to not pull in too many dependencies), I can surely implement the scanner as a function only.

On the other hand the constructor returns no error, so the call of the constructor + Scan function can be chained, so it doesn't really make a difference.

tokenscan/scanner.go
```diff
@@ -0,0 +1,221 @@
...
+			// synchronize start of token range scans, i.e. wait for all sessions to be created
+			sessionCreationWaitGroup.Done()
+			// Read ranges from the channel, until we have completed accessing all ranges
+			for r := range rangesChannel {
+				stmt, names := tokenSelect(selectBuilder, partitionKeyColumns...).ToCql()
+				q := gocqlx.Query(session.Query(stmt, names).WithContext(ctx), names).BindMap(qb.M{
+					"lower": r.startRange,
+					"upper": r.endRange,
+				})
+				iter := q.Iter()
```
mmatczuk, line 145: Compare that with https://github.com/scylladb/gocqlx/blob/a72115182914eced6612440463a3ab40c727c5b4/example_test.go#L586

tokenscan/scanner.go
mmatczuk: Single function is a better API, however, there can be an unexported scanner type to reduce worker and producer scope and to allow for code separation - unspaghetting :)

PR #164, opened 2020-06-02
qb.BatchBuilder: Remove deprecated flag and add comments on its limit…

GingerMoon:
…ation and warning

### DIFF
--- a/qb/batch.go
+++ b/qb/batch.go
@@ -23,6 +23,13 @@ type BatchBuilder struct {
 }
 
 // Batch returns a new BatchBuilder.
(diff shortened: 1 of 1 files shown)


mmatczuk:
Please fix linting, maybe change to single line comments.

GingerMoon:
Done.

GingerMoon:
@mmatczuk Any more comment? Should we keep the "Deprecated" flag?

mmatczuk:
We can drop the deprecated flag, I have no intention to remove it. In some cases it may work better.

GingerMoon:
OK. the Deprecated flag is added back.

PR #190, opened 2021-05-25
Correctly state in the GetQuery comment that it gets by primary key

### LINKED ISSUE #189: Table GetQuery() comment wrong
The comment of Table.GetQuery says:
```
// GetQuery returns query which gets by partition key.
```
This is in [this line](https://github.com/scylladb/gocqlx/blob/01be91c745c4563913e560ff27b7991aa411b149/table/table.go#L95).

However, this func calls `t.Get()`, which has the following comment:
```
// Get returns select by primary key statement.
```
(see [this line](https://github.com/scylladb/gocqlx/blob/01be91c745c4563913e560ff27b7991aa411b149/table/table.go#L83))

The latter is right, so the comment of the first func should be fixed so correctly speak of the `primary key`, not the `partition key`.

Alfa-Alex:
Fixes #189.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -92,7 +92,7 @@ func (t *Table) Get(columns ...string) (stmt string, names []string) {
 		ToCql()
 }
 
-// GetQuery returns query which gets by partition key.
+// GetQuery returns query which gets by primary key.
 func (t *Table) GetQuery(session gocqlx.Session, columns ...string) *gocqlx.Queryx {
 	return session.Query(t.Get(columns...))
 }


mmatczuk (approved):

mmatczuk:
Thanks!

alfa-alex:
I realized I forgot to fix the comment of the corresponding `GetQueryContext` function. Since this has been confusing me quite some times now, I propose to fix this. PR [here](https://github.com/scylladb/gocqlx/pull/257).

PR #206, opened 2021-11-20
queryx: unset empty values

### LINKED ISSUE #202: using of gocql.UnsetValue
Hi everyone! Recently we faced a following problem. In one of our processes we create a query by specifying ALL columns of a model struct and bind its fields by using method `BindStruct`. Some of the struct fields can be `nil` values and it leads to the problem that we generate ton of tombstones by explicitly sending `null` values to Cassandra. Base driver `gocql` has `gocql.UnsetValue` for unsetting value in statements that are already prepared. Unfortunately, creating custom type and adding method `gocql.MarshalCQL` doesn't solve the problem because `gocql` has different logic for set and unset values: we have to pass `gocql.UnsetValue` directly into `query.Bind` method.

To solve this problem we have some suggestions:
* Add method like `NilAsUnset` for `Queryx` struct. It sets some flag and then method `bindStructArgs` binds `gocql.UnsetValue` for all `nil` fields.
* Add interface with one method `BindCQL() interface{}`. Then in binding process check if value implements this interface. For these fields, instead of themselves, bind a return value of `BindCQL` method. As a result clients have flexibility in defining situations when custom fields should be considered as unset.

What do you think guys? Can we workaround stated problem without adding new functionality? If it's impossible, then we're ready to implement any of the suggested methods by ourselves.

N1cOs:
Resolve #202

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ ifndef GOTEST_CPU
 GOTEST_CPU := 1
 endif
 
+ifndef GOPATH
+GOPATH := $(shell go env GOPATH)
+endif
+
 ifndef GOBIN
 GOBIN := $(GOPATH)/bin
 endif
--- a/example_test.go
+++ b/example_test.go
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a ALv2-style
 // license that can be found in the LICENSE file.
 
+//go:build all || integration
 // +build all integration
 
 package gocqlx_test
--- a/example_test.go
+++ b/example_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/scylladb/gocqlx/v2/qb"
 	"github.com/scylladb/gocqlx/v2/table"
 	"golang.org/x/sync/errgroup"
+	"gopkg.in/inf.v0"
 )
 
 // Running examples locally:
@@ -47,6 +48,7 @@ func TestExample(t *testing.T) {
 	pagingEfficientFullTableScan(t, session)
 
 	lwtLock(t, session)
+	unsetEmptyValues(t, session)
 }
 
 // This example shows how to use query builders and table models to build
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@ GOTEST_CPU := 1
 endif
 
 ifndef GOBIN
-GOBIN := $(GOPATH)/bin
+GOBIN := $(shell go env GOPATH)/bin
```
mmatczuk, line 16: If missing, let's define GOPATH like GOBIN.

example_test.go
```diff
@@ -106,17 +107,30 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session) {
 
 	// Insert song using query builder.
 	insertSong := qb.Insert("examples.songs").
-		Columns("id", "title", "album", "artist", "tags", "data").Query(session)
```
mmatczuk, line 109: Please add new examples if appropriate.

transformer.go
```diff
@@ -0,0 +1,25 @@
...
+package gocqlx
+
+import (
+	"reflect"
+
+	"github.com/gocql/gocql"
+)
+
+// bindTransformer is called right before binding a value to a named parameter.
+type bindTransformer func(name string, val interface{}) interface{}
```
mmatczuk, line 14: Should be exported.

mmatczuk: I think we can leave is as `Transformer`, it looks generic enough to use it with parsing rows if needed.

transformer.go
```diff
@@ -0,0 +1,25 @@
...
+
+// UnsetEmpty unsets all empty parameters.
+// It helps to avoid tombstones when using the same insert/update
+// statement for filled and partially filled parameters.
+var UnsetEmpty = func(name string, val interface{}) interface{} {
```
mmatczuk, line 19: It should be more explicit that this is a transformer.

queryx.go
```diff
@@ -136,7 +144,7 @@ func (q *Queryx) BindStructMap(arg0 interface{}, arg1 map[string]interface{}) *Q
 	return q
 }
 
-func bindStructArgs(names []string, arg0 interface{}, arg1 map[string]interface{}, m *reflectx.Mapper) ([]interface{}, error) {
+func bindStructArgs(names []string, arg0 interface{}, arg1 map[string]interface{}, m *reflectx.Mapper, tr bindTransformer) ([]interface{}, error) {
```
mmatczuk, line 147: It already has 2 Query fields passed in let's make it a method.

transformer.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2017 ScyllaDB
+// Use of this source code is governed by a ALv2-style
+// license that can be found in the LICENSE file.
+
+package gocqlx
+
+// Transformer transforms the value of the named parameter to another value.
+type Transformer func(name string, val interface{}) interface{}
+
+// DefaultBindTransformer just do nothing.
```
N1cOs, line 16: It sounds a bit strange but I couldn't come up with anything better :)

Makefile
N1cOs: Done

example_test.go
N1cOs: I remained your insert statement, just save it in the separate variable. I suppose it's a good example how one statement can be used in situations with all filled params and partially filled.

If you think it's inappropriate in this example, I can remove it.

transformer.go
N1cOs: Named it as `Transformer`

transformer.go
N1cOs: Initially I though it would be some redundancy in word transformer but ok

queryx.go
N1cOs: Done

transformer.go
N1cOs: I set it as nil to save compatibility with older versions. Maybe someone uses insert/update statements with nil values to explicitly (or implicitly, who knows) delete columns

example_test.go
mmatczuk: It break the convention, feel free to copy the example and fully adjust it to present your usecase.

mmatczuk:
All is good let's fix the example. The current example is called `basicCreateAndPopulateKeyspace` please add something proper so that folks can grasp the idea.

example_test.go
N1cOs: Ok, added separate example

mmatczuk (approved):

mmatczuk:
Thanks for the contribution!

N1cOs:
@mmatczuk can we tag a new release, so we can update our dependencies?

mmatczuk:
Let's do that.

mmatczuk:
Here it is @N1cOs https://github.com/scylladb/gocqlx/releases/tag/v2.6.0

Issue #207, opened 2021-11-20
qb: placeholder for limit clause in select queries


N1cOs:
@mmatczuk What do you think about adding placeholder for `LIMIT` clause in select queries? I suppose it would be helpful, because now in one of our cases I have to concatenate strings to solve the issue.

N1cOs:
Oh, I decided to check for duplicates after sending it 😅 
Close it because it's a duplicate of #191

PR #208, opened 2021-11-20
qb: add named limit and per partition limit clauses

### LINKED ISSUE #191: Add named limit to QueryBuilder
Hi guys,
I was wondering if there is a possibility of adding an option to add named limit to query builder?
Currently, we can only use the `Limit` function and it injects the limit parameter directly into the statement.

N1cOs:
Resolve #191

### DIFF
--- a/qb/limit.go
+++ b/qb/limit.go
@@ -1,0 +1,45 @@
+// Copyright (C) 2017 ScyllaDB
+// Use of this source code is governed by a ALv2-style
+// license that can be found in the LICENSE file.
+
+package qb
+
+import (
+	"bytes"
+	"strconv"
+)
+
+type limit struct {
+	value        value
+	perPartition bool
+}
+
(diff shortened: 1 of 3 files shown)


mmatczuk:
Besides the new test cases what is the benefit of this change.

N1cOs:
> Besides the new test cases what is the benefit of this change.

What do you mean? I added two new methods: `LimitNamed` and `LimitPerPartitionNamed`. It resolves issue #191 and my duplicate issue #207

mmatczuk:
That's fine.

mmatczuk:
Thanks.

N1cOs:
By the way I also fixed the problem that `PER PARTITION LIMIT` clause must be before the `LIMIT` clause

mmatczuk:
That's excellent, thanks!

PR #209, opened 2021-12-10
Added test database name flag

spolischook:
```
go help testflag
```
> The 'go test' command may run tests for different packages
> 	    in parallel as well, according to the setting of the -p flag
> 	    (see 'go help build').

In parallel execution there will be a collision with single database name

### DIFF
--- a/gocqlxtest/gocqlxtest.go
+++ b/gocqlxtest/gocqlxtest.go
@@ -18,6 +18,7 @@ import (
 
 var (
 	flagCluster      = flag.String("cluster", "127.0.0.1", "a comma-separated list of host:port tuples")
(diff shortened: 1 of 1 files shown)


gocqlxtest/gocqlxtest.go
```diff
@@ -18,6 +18,7 @@ import (
 
 var (
 	flagCluster      = flag.String("cluster", "127.0.0.1", "a comma-separated list of host:port tuples")
+	flagDatabase     = flag.String("database", "gocqlx_test", "database name")
```
mmatczuk, line 21: This should be called key space

spolischook: good point! fixed

PR #218, opened 2022-02-10
Make qb.DESC of type qb.Order, not bool

Drahflow:
Accidentally discovered by code along the lines of:
https://play.golang.com/p/b_tcqJjkleI

Language spec says the const-declaration expression part repeats, but the type does not (there is even an example).


mmatczuk:
Thanks

PR #219, opened 2022-02-11
Add <Cmp>TupleNamed for two-sided range-queries on tuples

Drahflow:
I found no other way to 
```
sometable.SelectBuilder(colSomeCol).
    Where(qb.GtOrEqTupleNamed(colCreatedAtID, 2, "start")).
    Where(qb.LtTupleNamed(colCreatedAtID, 2, "end")).
```
i.e. do specify a range scan on a tuple column with both ends of the range given. With only `qb.GtOrEqTuple` and `qb.LtTuple` the autogenerated names are both derived from the column name and collide, thus it is not possible to specify different ends for the range.

Before merging this, please also consider that tuple-element syntax has some conflict with `gocql`, which produces `"col[0]"` in `SliceMap`, and not `"col_0"` as gocqlx wants it during tuple binding.

Some options I see:
* Rename this PR's functions (or add yet more) with `<Cmp>TupleIdx` resp. `<Cmp>TupleIdxNamed` which autogenerate `gocql` compatible column names.
* Accept API breakage :( and change `tupleparam`. 
* Introduce `<Cmp>TupleNamedElements(column string, names ...string)` which gives names for all elements (thereby also specifying count)

If you tell me which of these two options you'd like to see, I'll be preparing an according follow-up PR.

### DIFF
--- a/qb/cmp.go
+++ b/qb/cmp.go
@@ -90,6 +90,18 @@ func EqNamed(column, name string) Cmp {
 	}
 }
 
+// EqTupleNamed produces column=(?,?,...) with count number of placeholders and custom name.
+func EqTupleNamed(column string, count int, name string) Cmp {
+	return Cmp{
+		op:     eq,
+		column: column,
+		value: tupleParam{
+			param: param(name),
+			count: count,
+		},
+	}
+}
+
 // EqLit produces column=literal and does not add a parameter to the query.
 func EqLit(column, literal string) Cmp {
 	return Cmp{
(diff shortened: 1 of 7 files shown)


mmatczuk:
For the sake of completeness its fine to add the named variants to tuple compare functions - please add test cases.

The maximal benefit in this area would come from a dedicated API that would allow to operate on tuple filed level see example code below.

```go
type TupleCmp struct {
	column string
	lit   []lit
	param []param
}

func (t *TupleCmp) IdxParam(...) *TupleCmp {
}

func (t *TupleCmp) IdxParamNamed(...) *TupleCmp {
}

func (t *TupleCmp) IdxLit(...) *TupleCmp {
}

func (t *TupleCmp) Eq() Cmp {
	
}
...

func Tuple(column string) *TupleCmp {
	return &TupleCmp{
		column: column,
		lit:   make([]lit, 4),
		param: make([]param, 4),
	}
}
```

As for 

> Before merging this, please also consider that tuple-element syntax has some conflict with gocql, which produces "col[0]" in SliceMap, and not "col_0" as gocqlx wants it during tuple binding.

I think that it's not possible to bind a slice to a tuple, but maybe you can prove me wrong.

mmatczuk:
I guess we can fix that.

mmatczuk:
Thanks!

PR #220, opened 2022-02-15
avoid DATA RACE if reusing query builder

tehsphinx:
If calling `Query` or `QueryContext` on the same query builder in parallel there is a race condition in `using`.

This PR removes that race condition.

Example: query is built once, then reused:

```go
func getDelete(session gocqlx.Session) func(ctx context.Context, id string) error {
	qry := qb.Delete("tbl").Where(qb.Eq("id"))

	return func(ctx context.Context, id string) error {
		return qry.QueryContext(ctx, session).Bind(id).ExecRelease()
	}
}
```

### DIFF
--- a/qb/using.go
+++ b/qb/using.go
@@ -27,8 +27,6 @@ type using struct {
 	timestampName string
 	timeout       time.Duration
 	timeoutName   string
-
-	using bool
 }
 
 func (u *using) TTL(d time.Duration) *using {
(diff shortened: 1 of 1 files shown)


qb/using.go
```diff
@@ -71,46 +69,49 @@ func (u *using) TimeoutNamed(name string) *using {
...
 	}
 
 	return
 }
 
-func (u *using) writePreamble(cql *bytes.Buffer) {
-	if u.using {
-		cql.WriteString("AND ")
-	} else {
+func (u *using) getWritePreamble() func(cql *bytes.Buffer) {
```
mmatczuk, line 107: Can we rename that to preambleWriter

tehsphinx: absolutely 👍

tehsphinx:
@mmatczuk would it be possible to get a new version tag in the near future? That way we can pull this in without having to use a commit hash in all of our services or having to use a replace and create a version tag on our clone. That would be super helpful. 🙏

mmatczuk:
Here it is https://github.com/scylladb/gocqlx/releases/tag/v2.7.0

PR #229, opened 2022-03-23
fix using timeout order in select stmt

YeminLi:
Looking at the select statement specification [here](https://docs.scylladb.com/getting-started/dml/#select-statement),
it seems the USING clause should be at the end of the statement.


YeminLi:
Could you help to review this when having time? Thank you. @mmatczuk

tarzanek:
@avelanarius can we get a release for this?
at least a minor 2.7.1 or something?
we have people using this

PR #236, opened 2022-06-30
add user password authentication to schemagen

metaswirl:
This PR adds user and password flags to schemagen. We needed this to generate table models for a Cassandra/Scylla that is password-protected. I manually tested it.


mmatczuk:
Thanks.

PR #243, opened 2022-08-24
picks columns from table metadata for table.Update()

### LINKED ISSUE #242: Help Regarding table `Update`
I am using the `Update` like this.

```go
query := DB.Session.Query(user.GetTable().Update()).BindStruct(&user)
if err := query.ExecRelease(); err != nil {
	return err
}
```

But the error I am getting is .. `line 1:0 no viable alternative at input '<EOF>'`. I have tried `BindStruct` with struct as well as pointer. Same error. Attaching the debugger, there are no values. Where might I be going wrong. Logging the query statement gives me

`[query statement=\"UPDATE github_installations SET  \" values=[] consistency=QUORUM]`

Screen Shot 2022-08-24 at 9 52 49 AM

debuggerpk:
This is an attempt to fix #242. Work in progress.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -157,6 +157,9 @@ func (t *Table) InsertBuilder() *qb.InsertBuilder {
 
 // Update returns update by primary key statement.
 func (t *Table) Update(columns ...string) (stmt string, names []string) {
+	if len(columns) == 0 {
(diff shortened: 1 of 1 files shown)


debuggerpk:
This should work, but Now I am getting an error `line 1:0  : syntax error...\n` Any help would be appreciated.

Issue #247, opened 2023-01-30
gocqlxtest: get test server configuration as from flags so from the environment variables


Ivan-Feofanov:
Hi guys.

gocqlxtest is very useful but I want to have the ability to take my test server config not only from flags but also from env variables, it extends flexibility for using it, for example in CI pipelines.

dkropachev:
It is better just copy paste that code, than having unstable dependancy.

PR #248, opened 2023-01-30
gocqlxtest: get config from env

Ivan-Feofanov:
PR for the issue #247 
With the replacement builtin `flags` package to `github.com/namsral/flag` we could take config as from flags so from env variables

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,26 @@ make run-scylla
 make run-examples
 ```
 
+## Testing
+
+For the testing purpose you can use `gocqlxtest` package:
+```go
+session := gocqlxtest.CreateSession(t)
+```
+To configure test session connection you can use both command-line
+arguments and environment variables:
+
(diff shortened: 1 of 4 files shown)


Gor027:
@avelanarius LGTM. Can you merge it?

Gor027 (approved):
@avelanarius LGTM. Can you merge it?

zimnx:
@Gor027 I don't think such little feature requires additional dependency (tagged as `pre`) to entire project. Keeping low subset of dependencies in libraries is a good thing. I'm against merging this.

mmatczuk:
It's not worth adding another dependency for all users.

Having said that the Gocqlx core module could add some automation around env vars but it needs to be explicit (well defined function) and not use 3rd party libraries.

dkropachev:
Agree, I recommend just copy paste that code, it is not more than 50 lines in total

PR #250, opened 2023-02-02
Support passing in custom columns for insert stmt

YeminLi:
support insert into partial columns (not all).


zimnx:
Please squash commits before merge, and explain what these changes are brining it commit description.

PR #251, opened 2023-02-03
expose SelectBuilder for table Get operation

YeminLi:
Hi,

In this PR, we add a new function to return the SelectBuilder for the table Get operation.

Thank you


dkropachev:
Thanks @YeminLi

PR #255, opened 2023-05-30
added timeout flag for the cassandra server connection timeout

SandeepDev1:
Just added the timeout flag for the cluster.ConnectTimeout
I needed for my server so thought it can be useful for others too.

### DIFF
--- a/cmd/schemagen/schemagen.go
+++ b/cmd/schemagen/schemagen.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/gocql/gocql"
 	"github.com/scylladb/gocqlx/v2"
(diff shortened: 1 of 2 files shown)


cmd/schemagen/schemagen.go
```diff
@@ -26,6 +27,7 @@ var (
 	flagOutput   = cmd.String("output", "models", "the name of the folder to output to")
 	flagUser     = cmd.String("user", "", "user for password authentication")
 	flagPassword = cmd.String("password", "", "password for password authentication")
+	flagTimeout  = cmd.Int("timeout", 5, "connection timeout for the connection to the database ( in seconds )")
```
sylwiaszunejko, line 30: I think default should be in line with `gocql` default value, which is `11` (https://github.com/gocql/gocql/blob/master/cluster.go#L262)

sylwiaszunejko (changes requested):

dkropachev:
Closed in favor of https://github.com/scylladb/gocqlx/pull/293

PR #269, opened 2024-06-05
Marshal missing UDT fields as null instead of failing

sylwiaszunejko:
We can't return an error in case a field is added to the UDT, otherwise existing code would break by simply altering the UDT in the database. For extra fields at the end of the UDT put nulls to be in line with gocql, but also python-driver and java-driver.

In gocql it was fixed in https://github.com/scylladb/gocql/commit/d2ed1bb74f3118a83a352e9ce912be765001efa4

java-driver https://github.com/datastax/java-driver/blob/ef56d561d97adcae48e0e6e8807f334aedc0d783/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/UdtCodec.java#L86
python-driver: https://github.com/datastax/python-driver/blob/15d715f4e686032b02ce785eca1d176d2b25e32b/cassandra/cqltypes.py#L1036

### DIFF
--- a/iterx.go
+++ b/iterx.go
@@ -13,7 +13,7 @@ import (
 	"github.com/scylladb/go-reflectx"
 )
 
-// DefaultUnsafe enables the behavior of forcing the iterator to ignore
+// DefaultUnsafe enables the behavior of forcing queries and iterators to ignore
 // missing fields for all queries. See Unsafe below for more information.
 var DefaultUnsafe bool
 
--- a/iterx_test.go
+++ b/iterx_test.go
@@ -9,6 +9,7 @@ package gocqlx_test
 
 import (
 	"math/big"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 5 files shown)


udt.go
```diff
@@ -39,11 +39,18 @@ func makeUDT(value reflect.Value, mapper *reflectx.Mapper, unsafe bool) udt {
 
 func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
 	value, ok := u.field[name]
-	if !ok {
-		return nil, fmt.Errorf("missing name %q in %s", name, u.value.Type())
+
+	var data []byte
+	var err error
+	if ok {
+		var err error
```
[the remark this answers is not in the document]
sylwiaszunejko: my bad, fixed it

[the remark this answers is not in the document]
mykaul: nit: I wonder if we can add https://github.com/dominikh/go-tools to CI here.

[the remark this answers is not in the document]
sylwiaszunejko: we can try, should I create a separate issue for that?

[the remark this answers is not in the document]
mykaul: of course.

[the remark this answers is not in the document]
sylwiaszunejko: done

sylwiaszunejko:
@avelanarius ping

udt.go
```diff
@@ -39,11 +39,17 @@ func makeUDT(value reflect.Value, mapper *reflectx.Mapper, unsafe bool) udt {
 
 func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
 	value, ok := u.field[name]
-	if !ok {
-		return nil, fmt.Errorf("missing name %q in %s", name, u.value.Type())
+
```
dkropachev, line 42: Should we consider `unsafe` flag ?
```
func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
	value, ok := u.field[name]
	if ok {
		return gocql.Marshal(info, value.Interface())
	}
	if u.unsafe {
		return nil, nil
	}
	return nil, fmt.Errorf("missing name %q in %s", name, u.value.Type())
}
```

dkropachev (commented):
We definitely should have test for that

udt.go
sylwiaszunejko: @dkropachev AFAIK this would require a user to use `query.Iter().Unsafe()` instead of just `query.Exec()`, because in `gocql` implementation there is no `unsafe` equivalent. Wouldn't it be counterintuitive if in `gocql` we are able to do `query.Exec()` with udt with missing fields and it works, but in `gocqlx` we are required to use `Unsafe()`?

dkropachev: @sylwiaszunejko , 100% agree on that, but since Unsafe is there and does control this particular behavior we have to be consistent in the way it works.

dkropachev:
> To sum up after a discussion with @dkropachev: There is a mechanism in `gocqlx` for ignoring missing UDT fields, but it requires a user to use `query.Iter().Unsafe()` instead of just `query.Exec()`, because in `gocql` implementation there is no `Unsafe` equivalent. So on one hand `Unsafe` is there and does control this particular behavior we have to be consistent in the way it works, but on the other hand we have to be consistent with `gocql` and if using UDTs with extra fields is supported there always not only when extra option specified, shouldn't it be also supported in `gocqlx`? @mykaul do you have any thoughts on that? Because it looks like my fix doesn't take `Unsafe` mechanism in the consideration and that's probably not the best

I would propose to rename this attribute to `Strict` reversing it's meaning, making default to be false.
Only problem here is that we have to release new major for that, due to the compatibility issue.

mykaul:
I'm in favor of consistency, even if it's the longer path (and complexity). The unsafe is not intuitive.

dkropachev:
@mykaul , could you please rereview it, tests were added and more cases fixed.
Also new method was introduced to `Queryx` to control `Unsafe` flag on it's level instead of borrowing it from `DefaultUnsafe`

mykaul:
Do we have an example (in gocqlx docs) for UDT usage?

sylwiaszunejko:
@mykaul  examples are here: https://github.com/scylladb/gocqlx/blob/a9ab2701967cba997c7c833ad3362f0659438228/doc_test.go#L24 and https://github.com/scylladb/gocqlx/blob/a9ab2701967cba997c7c833ad3362f0659438228/example_test.go#L445

iterx.go
```diff
@@ -206,9 +206,9 @@ func (iter *Iterx) scanAll(dest interface{}) bool {
 
 // isScannable takes the reflect.Type and the actual dest value and returns
 // whether or not it's Scannable. t is scannable if:
-//   * ptr to t implements gocql.Unmarshaler, gocql.UDTUnmarshaler or UDT
-//   * it is not a struct
-//   * it has no exported fields
+//   - ptr to t implements gocql.Unmarshaler, gocql.UDTUnmarshaler or UDT
+//   - it is not a struct
+//   - it has no exported fields
```
wprzytula, line 211: Why this change? Asterisks are valid Markdown itemizers as well AFAIK.

iterx_test.go
```diff
@@ -480,12 +480,8 @@ func TestIterxUnsafe(t *testing.T) {
...
-		gocqlx.DefaultUnsafe = true
-		defer func() {
-			gocqlx.DefaultUnsafe = false
-		}()
 		var v []UnsafeTable
-		err := session.Query(stmt, nil).Iter().Select(&v)
+		err := session.Query(stmt, nil).Unsafe().Iter().Select(&v)
 		if err != nil {
 			t.Fatal("Select() failed:", err)
 		}
```
wprzytula, line 808: What was the purpose of the removed code with `defer` and why can we remove it now?

wprzytula: Ah, is it because we migrate `DefaultUnsafe` from being a global gocqlx attribute to be per-query attribute?

queryx.go
```diff
@@ -216,6 +216,13 @@ func (q *Queryx) Bind(v ...interface{}) *Queryx {
 	return q
 }
 
+// Scan executes the query, copies the columns of the first selected
+// row into the values pointed at by dest and discards the rest. If no rows
+// were selected, ErrNotFound is returned.
+func (q *Queryx) Scan(v ...interface{}) error {
+	return q.Query.Scan(udtWrapSlice(q.Mapper, q.unsafe, v)...)
+}
+
```
wprzytula, line 225: Wow, do we introduce a new public API in this commit? If so, I'd much prefer to have it sent as a distinct commit, well described what it is and what for.

udt.go
```diff
@@ -39,26 +39,24 @@ func makeUDT(value reflect.Value, mapper *reflectx.Mapper, unsafe bool) udt {
...
-		return fmt.Errorf("missing name %q in %s", name, u.value.Type())
+	if ok {
+		return gocql.Unmarshal(info, data, value.Addr().Interface())
 	}
-
-	return gocql.Unmarshal(info, data, value.Addr().Interface())
+	if u.unsafe {
+		return nil
+	}
+	return fmt.Errorf("missing name %q in %s", name, u.value.Type())
```
wprzytula, line 59: Are we sure this is the semantics of `unsafe` that we want to have?
In Rust driver at least, we consider _too many_ and _too few_ fields in the CQL UDT compared to the go struct as two distinct cases. In Rust driver, _strict_ forbids _too many_ fields in CQL UDT, whereas _default_if_missing_ is used for the case when a struct has a field that is not found in the CQL UDT metadata. WDYT? Should we cover it by a single lever here?

wprzytula:
I'm strongly for the change from `unsafe` to `strict`, and making non-`strict` the default. This would be a breaking change, but one that I find justified.

sylwiaszunejko:
> I'm strongly for the change from `unsafe` to `strict`, and making non-`strict` the default. This would be a breaking change, but one that I find justified.

I agree, I would like to go with the `strict` approach.

@mykaul @avelanarius @roydahan Are we okay with making breaking change in that case?

iterx_test.go
dkropachev: > Ah, is it because we migrate `DefaultUnsafe` from being a global gocqlx attribute to be per-query attribute?

Exactly

queryx.go
dkropachev: It might seemse like we do, but we don't, since `Queryx` has inherits all public methods from `Query` it was there before, but due to lack of wrapping was not behaving in the same was as other public methods of `Queryx`.

udt.go
dkropachev: @wprzytula, i am totally in support of having all drivers as close as possible to each other. Let's have separate issue on that and address it in other PR

queryx.go
wprzytula: I'm not sure if I got it. You mean that it was previously possible to call `Queryx.Scan`, and it was just `Query.Scan` - without calling `udtWrapSlice` on passed arguments?

udt.go
wprzytula: OTOH I have no idea how other drivers apart from Rust driver approach this.

queryx.go
dkropachev: correct

iterx.go
dkropachev: This one is gone, it is already in master, reason why it have changed is `golangci-lint` does not tollerate anything by `-` there.

queryx.go
wprzytula: Then what you add here is not introducing new public function, but rather fixing an existing one. OK!

udt.go
dkropachev: We can start from these two

dkropachev:
> > I'm strongly for the change from `unsafe` to `strict`, and making non-`strict` the default. This would be a breaking change, but one that I find justified.
> 
> I agree, I would like to go with the `strict` approach.
> 
> @mykaul @avelanarius @roydahan Are we okay with making breaking change in that case?

Let's do that in separate PR.

roydahan:
I'm fine with it, we can release a major and state that in the release notes.

sylwiaszunejko:
I will go with the strict approach in a separate PR

sylwiaszunejko:
Let's merge it and then we can replace unsafe with strict approach in the additional PR. Are you okay with that?
@wprzytula @roydahan

wprzytula (approved):

mmatczuk:
Interesting.

mmatczuk:
I'd refrain from bumping the major version and only do that in case of major API breakage i.e. new API, preferably never. 

Bumping it to v2 was a complete mess, every import line needs to be changed.
Not that it's hard, but it puts quite a lot of annoying work on customers.

sylwiaszunejko:
@mmatczuk do you have any suggestion how to handle better the situation with Unsafe and how the default behavior around dealing with missing fields in UDTs differs in gocql and gocqlx?

mmatczuk:
The change is LGTM.

sylwiaszunejko:
I am not sure if I understand right, is it okay to do breaking change without major release?

mmatczuk:
What I'm trying to say is that there is a cost of making /v3 it is not something to do lightly.
Consider what you need, maybe completely removing unsafe and releasing /v3 is a way to go.

dkropachev:
> What I'm trying to say is that there is a cost of making /v3 it is not something to do lightly. Consider what you need, maybe completely removing unsafe and releasing /v3 is a way to go.

Just want to point out that gocqlx recently (https://github.com/scylladb/gocqlx/pull/274), which also contributes to necessity to bump major up.

PR #273, opened 2024-06-14
Update golangci-lint in Makefile and CI and turn it on in CI

dkropachev:
No functional changes.
Update `golangci-lint`, add it to github workflow, address all linter complaints.

### DIFF
--- a/batchx.go
+++ b/batchx.go
@@ -4,6 +4,7 @@ import (
 	"github.com/gocql/gocql"
 )
 
+// Batch is a wrapper around gocql.Batch
 type Batch struct {
 	*gocql.Batch
 }
--- a/cmd/schemagen/schemagen.go
+++ b/cmd/schemagen/schemagen.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/gocql/gocql"
+
 	"github.com/scylladb/gocqlx/v2"
 	_ "github.com/scylladb/gocqlx/v2/table"
 )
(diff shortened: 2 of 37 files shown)


iterx.go
```diff
@@ -19,17 +19,14 @@ var DefaultUnsafe bool
...
+	fields     [][]int
+	values     []interface{}
 	unsafe     bool
 	structOnly bool
 	applied    bool
-	err        error
-
-	// Cache memory for a rows during iteration in structScan.
-	fields [][]int
-	values []interface{}
```
sylwiaszunejko, line 32: Are we okay with losing the information that this comment gives?

batchx.go
```diff
@@ -4,6 +4,7 @@ import (
 	"github.com/gocql/gocql"
 )
 
+// Batch a batch operations.
```
sylwiaszunejko, line 7: I don't quite get this comment

cmd/schemagen/schemagen.go
```diff
@@ -169,7 +167,7 @@ func existsInSlice(s []string, v string) bool {
 // The second element contains the name of the type.
 //
 //	[["<my_type,", "my_type"] ["my_other_type>", "my_other_type"]]
-var userTypes = regexp.MustCompile(`(?:<|\s)(\w+)(?:>|,)`) // match all types contained in set<X>, list<X>, tuple<A, B> etc.
+var userTypes = regexp.MustCompile(`(?:<|\s)(\w+)[>,]`) // match all types contained in set<X>, list<X>, tuple<A, B> etc.
```
sylwiaszunejko, line 170: why this change?

dkropachev: Regexp simplification, one of the linter does that

iterx.go
dkropachev: thanks for pointing out, fixed.

batchx.go
dkropachev: Added something that makes more sense

sylwiaszunejko (approved):

PR #276, opened 2024-06-25
Replace Unsafe with Strict mechanism

sylwiaszunejko:
Previously by default the presence of a missing field in a udt would result in an error reported. The Unsafe mechanism could be used to ignore these fields.

This PR changes the default behavior to ignoring missing fields and only reporting an error if Strict mode is used. This approach is in line with the gocql. 

This is a breaking change and the major release will be needed.

Refs: #269


dkropachev (approved):
Great job

wprzytula (approved):

PR #277, opened 2024-06-26
Switch to use scylla image instead of scylla version, to be able to target enterprise

dkropachev:
Currently `Makefile` uses `SCYLLA_VERSION` to define docker image.
Switching to `SCYLLA_IMAGE` instead will help to target enterprise, nightly or custom image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     env:
-      SCYLLA_VERSION: 6.0.0
+      SCYLLA_IMAGE: scylladb/scylla:6.0.0
       GOBIN: ./bin
     steps:
       - name: Git Checkout
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 all: check test
 
-ifndef SCYLLA_VERSION
-SCYLLA_VERSION := latest
+ifndef SCYLLA_IMAGE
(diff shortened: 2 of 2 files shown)


sylwiaszunejko:
I am okay with merging this commit: [Switch to use scylla image instead of scylla version, to be able to target enterprise](https://github.com/scylladb/gocqlx/pull/277/commits/76853189196011c3544d6100372929902d2f0c2d)
I would just suggest change to the commit message to something like that:
```
Switch to use scylla image instead of scylla version

Currently Makefile uses SCYLLA_VERSION to define docker image.
Switching to SCYLLA_IMAGE instead will help to target enterprise, nightly or custom image
```

but the second commit shouldn't be in this PR

dkropachev:
> I am okay with merging this commit: [Switch to use scylla image instead of scylla version, to be able to target enterprise](https://github.com/scylladb/gocqlx/pull/277/commits/76853189196011c3544d6100372929902d2f0c2d) I would just suggest change to the commit message to something like that:
> 
> ```
> Switch to use scylla image instead of scylla version
> 
> Currently Makefile uses SCYLLA_VERSION to define docker image.
> Switching to SCYLLA_IMAGE instead will help to target enterprise, nightly or custom image
> ```
> 
> but the second commit shouldn't be in this PR

Commit leaked to master and I did not spot it right away, thanks, fixed

sylwiaszunejko (approved):

PR #278, opened 2024-06-26
Make schemagen generate compilable code and add test for it

dkropachev:
It turns out that `schemagen` does not generate compilable file after switching from `v1` to `v2`
Make it compilable and have a test for it


cmd/schemagen/schemagen_test.go
```diff
@@ -30,7 +30,7 @@ func TestSchemagen(t *testing.T) {
 
 	b := runSchemagen(t, "foobar")
 
-	const goldenFile = "testdata/models.go.txt"
```
sylwiaszunejko, line 33: What's the reason for a file name change?

cmd/schemagen/schemagen.go
```diff
@@ -120,8 +124,8 @@ func renderTemplate(md *gocql.KeyspaceMetadata) ([]byte, error) {
 			if c.Type == "decimal" && !existsInSlice(imports, "gopkg.in/inf.v0") {
 				imports = append(imports, "gopkg.in/inf.v0")
 			}
-			if c.Type == "duration" && !existsInSlice(imports, "github.com/gocql/gocql") {
-				imports = append(imports, "github.com/gocql/gocql")
```
sylwiaszunejko, line 124: Why adding `v2` to gocql? I thought we add `v2` to `gocqlx` not `gocql`.

cmd/schemagen/schemagen_test.go
dkropachev: It has changed since I want to test generated code for real

cmd/schemagen/schemagen.go
dkropachev: Fixed, thanks.

.gitignore
```diff
@@ -15,4 +15,5 @@ vendor
 # GoLand IDE
 .idea/
 
-cmd/schemagen/schemagen
\ No newline at end of file
+cmd/schemagen/schemagen
+cmd/schemagen/testdata/go.sum
```
sylwiaszunejko, line 19: Lack of a newline

cmd/schemagen/testdata/models_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gocql/gocql"
+	"github.com/scylladb/gocqlx/v2"
+	"github.com/scylladb/gocqlx/v2/qb"
+)
+
+var flagCluster = flag.String("cluster", "127.0.0.1", "a comma-separated list of host:port tuples")
```
sylwiaszunejko, line 15: comment states "host:port" but the default is only "host"

cmd/schemagen/schemagen.go
```diff
@@ -109,12 +109,16 @@ func renderTemplate(md *gocql.KeyspaceMetadata) ([]byte, error) {
...
+	if len(md.Types) != 0 {
+		imports = append(imports, "github.com/scylladb/gocqlx/v2")
+	}
+
 	for _, t := range md.Tables {
 		// Ensure ordered columns are sorted alphabetically
 		sort.Strings(t.OrderedColumns)
 
 		for _, c := range t.Columns {
-			if (c.Type == "timestamp" || c.Type == "date" || c.Type == "duration" || c.Type == "time") && !existsInSlice(imports, "time") {
```
sylwiaszunejko, line 117: Why removing "duration" type here?

.gitignore
dkropachev: Added, thankls

cmd/schemagen/testdata/models_test.go
dkropachev: Fixed

cmd/schemagen/schemagen.go
dkropachev: i think it was leftover since time when duration was represented as time.Duration, as of now is getting converted into `gocql.Duration` and therefore no `time` package needed for it.

PR #286, opened 2024-08-29
feat(migrate): add `Pending` function for pending migrations

JustSamuel:
This PR adds a `Pending` function to the `package migration`.

 Can be used to ensure that an application is up to date before launching.

### DIFF
--- a/migrate/migrate.go
+++ b/migrate/migrate.go
@@ -87,6 +87,48 @@ func List(ctx context.Context, session gocqlx.Session) ([]*Info, error) {
 	return v, nil
 }
 
+// Pending provides a listing of pending migrations.
+func Pending(ctx context.Context, session gocqlx.Session, f fs.FS) ([]*Info, error) {
+	applied, err := List(ctx, session)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a set of applied migration names
+	appliedNames := make(map[string]struct{}, len(applied))
+	for _, migration := range applied {
+		appliedNames[migration.Name] = struct{}{}
+	}
+
+	fm, err := fs.Glob(f, "*.cql")
+	if err != nil {
+		return nil, fmt.Errorf("list migrations: %w", err)
+	}
+
+	pending := make([]*Info, 0)
+
+	for _, name := range fm {
+		baseName := filepath.Base(name)
+		// Check if the migration is not in the applied set
+		if _, exists := appliedNames[baseName]; !exists {
+			c, err := fileChecksum(f, name)
+			if err != nil {
+				return nil, fmt.Errorf("calculate checksum for %q: %w", name, err)
+			}
+
+			info := &Info{
+				Name:      baseName,
+				StartTime: time.Now(),
+				Checksum:  c,
+			}
+
+			pending = append(pending, info)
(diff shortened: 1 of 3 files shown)


sylwiaszunejko:
Pls address CI failures and it would be nice to have tests/example of usage

JustSamuel:
> Pls address CI failures and it would be nice to have tests/example of usage

Fixed

sylwiaszunejko:
CI still fails

JustSamuel:
Should be fine now

migrate/migrate.go
```diff
@@ -87,6 +87,48 @@ func List(ctx context.Context, session gocqlx.Session) ([]*Info, error) {
...
+		return nil, err
+	}
+
+	fm, err := fs.Glob(f, "*.cql")
+	if err != nil {
+		return nil, fmt.Errorf("list migrations: %w", err)
+	}
+
+	if len(applied) > len(fm) {
+		return nil, fmt.Errorf("database is ahead")
```
dkropachev, line 103: no need to have validation here, if it is called asynchronously you will get into troubles.
If you want to validate it, just add another method `ValidateFS(ctx context.Context, session gocqlx.Session, f fs.FS) error` and do just that there.

migrate/migrate.go
```diff
@@ -87,6 +87,48 @@ func List(ctx context.Context, session gocqlx.Session) ([]*Info, error) {
...
+	}
+
+	pending := make([]*Info, 0, len(fm)-len(applied))
+
+	for i := range applied {
+		if applied[i].Name != fm[i] {
+			return nil, fmt.Errorf("inconsistent migrations found, expected %q got %q at %d", applied[i].Name, fm[i], i)
```
dkropachev, line 110: Same here.

migrate/migrate.go
```diff
@@ -87,6 +87,48 @@ func List(ctx context.Context, session gocqlx.Session) ([]*Info, error) {
...
+		}
+	}
+
+	for _, name := range fm[len(applied):] {
+		c, err := fileChecksum(f, name)
+		if err != nil {
+			return nil, fmt.Errorf("calculate checksum for %q: %w", name, err)
```
dkropachev, line 117: Same here.

dkropachev (changes requested):

migrate/migrate.go
```diff
@@ -87,6 +87,48 @@ func List(ctx context.Context, session gocqlx.Session) ([]*Info, error) {
...
+
+	pending := make([]*Info, 0, len(fm)-len(applied))
+
+	for i := range applied {
+		if applied[i].Name != fm[i] {
+			return nil, fmt.Errorf("inconsistent migrations found, expected %q got %q at %d", applied[i].Name, fm[i], i)
+		}
+	}
+
+	for _, name := range fm[len(applied):] {
```
dkropachev, line 114: It is better to just to go over `fm` and select only those that are not present in `applied` by `.Name`

JustSamuel: I put all the `applied`  in a set and check over the `fs` to find the difference.

dkropachev (approved):

PR #290, opened 2024-09-10
Add option to add ALLOW FILTERING to the update query

sylwiaszunejko:
Previously it was only possible to use `AllowFiltering()` on the select query builder.

This PR adds the option to add `ALLOW FILTERING` to the update query.

Fixes: #288

### DIFF
--- a/qb/update.go
+++ b/qb/update.go
@@ -37,6 +37,7 @@ type UpdateBuilder struct {
 	_if         _if
 	using       using
 	exists      bool
+	allowFiltering    bool
 }
 
 // Update returns a new UpdateBuilder with the given table name.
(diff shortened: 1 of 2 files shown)


sylwiaszunejko:
@dkropachev should different types of queries also have this option?

dkropachev (approved):

dkropachev:
> @dkropachev should different types of queries also have this option?

yes, `select` as well

sylwiaszunejko:
@dkropachev  select already has it

dkropachev:
> @dkropachev select already has it, any other?

Then it is all good

kennethy:
Just to confirm if `allow filtering` is actually supported for update queries.

I don't think it is as per the [official docs](https://opensource.docs.scylladb.com/master/cql/dml/update.html#). 🤔

dkropachev:
> Just to confirm if `allow filtering` is actually supported for update queries.
> 
> I don't think it is as per the [official docs](https://opensource.docs.scylladb.com/master/cql/dml/update.html#). 🤔

It is not.

Issue #322, opened 2025-06-04
Batch in gocqlx doesn't support adding context.


akshat1998:
Batch in gocqlx, created using session.NewBatch(), doesn't support adding context.
Although we can add context to individual queries, like below:
```
	batch := c.session.NewBatch(gocql.LoggedBatch)
	stmt, names := qb.Insert(query.TableName).ToCql()
	queryX := c.session.Query(stmt, names).WithContext(ctx)
	err := batch.BindStruct(queryX, source)
	if err != nil {
		return err
	}
```

So any reason why this was avoided, or this is just missing as of now?

UPD: query context is lost when adding query to batch. [Ref](https://github.com/scylladb/gocqlx/issues/322#issuecomment-2961114267).

dkropachev:
> Batch in gocqlx, created using session.NewBatch(), doesn't support adding context. Although we can add context to individual queries, like below:
> 
> ```
> 	batch := c.session.NewBatch(gocql.LoggedBatch)
> 	stmt, names := qb.Insert(query.TableName).ToCql()
> 	queryX := c.session.Query(stmt, names).WithContext(ctx)
> 	err := batch.BindStruct(queryX, source)
> 	if err != nil {
> 		return err
> 	}
> ```
> 
> So any reason why this was avoided, or this is just missing as of now?

query context is lost when query is added to the batch.

First we will merge gocql part: https://github.com/scylladb/gocql/pull/501, then we add it here.

dkropachev:
> query context is lost when query is added to the batch.
> 
> First we will merge gocql part: [scylladb/gocql#501](https://github.com/scylladb/gocql/pull/501), then we add it here.

Let's don't way optimization, I have created PR - https://github.com/scylladb/gocqlx/pull/326 to add API

akshat1998:
Thank you @dkropachev for the quick PR. 🙌

PR #326, opened 2025-06-11
Add ContextBatch api to create batch with context

dkropachev:
Fixes: https://github.com/scylladb/gocqlx/issues/322

Adds missing `ContextBatch` and `Batch` API, marks `NewBatch` as depricated.

### DIFF
--- a/batchx.go
+++ b/batchx.go
@@ -1,8 +1,8 @@
 package gocqlx
 
 import (
+	"context"
 	"fmt"
-
 	"github.com/gocql/gocql"
 )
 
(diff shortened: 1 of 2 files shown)


batchx.go
```diff
@@ -12,9 +12,25 @@ type Batch struct {
 }
 
 // NewBatch creates a new batch operation using defaults defined in the cluster.
+//
+// Deprecated: use session.Batch instead
```
akshat1998, line 16: qb.Batch() has redirection to this.
> // Deprecated: Please use gocql.Session.NewBatch() instead.
So would need to change there as well.

akshat1998:
Hey @dkropachev ,
Why are we not adding context like this:
```
func (b *Batch) WithContext(ctx gocql.Context) *Batch {
	b.Batch = b.Batch.WithContext(ctx)
	return b
}
```

akshat1998:
Ok, got the reasoning for above.
This PRs description: https://github.com/scylladb/gocql/pull/501

dkropachev:
To have same behavior as gocql.API that returns copy of Batch

batchx.go
```diff
@@ -12,9 +12,25 @@ type Batch struct {
...
+
+// Batch creates a new batch operation using defaults defined in the cluster.
+func (s *Session) Batch(bt gocql.BatchType) *Batch {
+	return &Batch{
+		Batch: s.Session.Batch(bt),
+	}
+}
+
+// ContextBatch creates a new batch operation using defaults defined in the cluster with attached context.
+func (s *Session) ContextBatch(ctx context.Context, bt gocql.BatchType) *Batch {
```
akshat1998, line 31: Better to use similar naming as [here](https://github.com/scylladb/gocql/pull/501/files#diff-3d4a6b7d54b5107bc5694b7a06383f9ec4c68f9abe43ba0942f5d970f3a8ccacR2025).
`BatchWithContext` is used there.

dkropachev: Thanks for taking time to review.
Agree with you on that.
Unfortunately, Query API is already there, we need to maintain uniformity.

Let's do this, I will add this API, when that or is merged and we update gocql, I will introduce new API and mark old one as deprecated.

Issue #330, opened 2025-07-03
3.0.2 introduces a breaking change


max-melentyev:
Hey, I'm getting errors after updating 3.0.1 -> 3.0.2

```
../../go/pkg/mod/github.com/scylladb/gocqlx/v3@v3.0.2/batchx.go:26:10: b.Batch.SetHostID undefined (type *gocql.Batch has no field or method SetHostID)
../../go/pkg/mod/github.com/scylladb/gocqlx/v3@v3.0.2/queryx.go:155:10: q.Query.SetHostID undefined (type *gocql.Query has no field or method SetHostID)
```

Should this tag be deleted and released as v4, as it has a breaking change?

max-melentyev:
Yes, I'm sure. gocqlx [requires gocql 1.7.0](https://github.com/scylladb/gocqlx/blob/master/go.mod#L6), the version you bumped is in the `replace` instruction which only applies to the gocqlx repo, and it's not used to resolve dependencies of projects that use gocqlx.

sylwiaszunejko:
@max-melentyev Please see https://github.com/scylladb/gocqlx/pull/331#issuecomment-3044607957, from gocqlx v3.0.0 we use scylladb fork of gocql, and as for now there is no other way to make it work then to use `replace` directive and we don't have any way to enforce this requirement using `require`

max-melentyev:
Sounds good, does it make sense to add [retract](https://go.dev/ref/mod#go-mod-file-retract) for 3.0.2 because it has broken dependencies?

dkropachev:
> Sounds good, does it make sense to add [retract](https://go.dev/ref/mod#go-mod-file-retract) for 3.0.2 because it has broken dependencies?

It doesn't have broken dependency, if you add proper replace it would work just fine, it used to be that way from the very beginning.

dkropachev:
> 3.0.1 worked without replace. 3.0.2 doesn't work. This is a breaking change, and it requires major version bump. Otherwise `go get` and renovate bot will be using 3.0.2 where 3.0.1 is used and this will not work apparently.

It is misuse, `scylladb/gocqlx` meant to be used alongside with `scylladb/gocql` only.

max-melentyev:
> there is no other way to use it with scylladb/gocql

There is a way, use `github.com/scylladb/gocql` instead of `github.com/gocql/gocql`, without replace. 

> you have to provide replace directive

[...]

PR #331, opened 2025-07-07
Bump required gocql version

sylwiaszunejko:
`SetHostID` methods added in gocql v1.15.1 are needed for new API added in 3.0.2. 

Refs: #330

### DIFF
--- a/cmd/schemagen/testdata/go.mod
+++ b/cmd/schemagen/testdata/go.mod
@@ -3,7 +3,7 @@ module schemagentest
 go 1.20
 
 require (
-	github.com/gocql/gocql v1.7.0
+	github.com/gocql/gocql v1.15.1
 	github.com/google/go-cmp v0.7.0
 	github.com/scylladb/gocqlx/v3 v3.0.0
 )
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/scylladb/gocqlx/v3
 go 1.20
 
 require (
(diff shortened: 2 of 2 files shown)


sylwiaszunejko:
> Did you tested it ?
> 
> I gave it a test and found that it does not work.
> 
> golang [manuals](https://go.dev/doc/modules/gomod-ref#replace-notes) reads:
> 
> ```
> Use the exclude and replace directives to control build-time dependency resolution when building the current module. These directives are ignored in modules that depend on the current module.
> ```
> 
> Which means that when you upgrade `gocqlx` you need to upgrade `github.com/scylladb/gocql` version in the replace directive manually.

I see, I kinda forgot that in this require there is gocql/gocql no scylladb/gocql, can we somehow require scylladb version not the upstream one? Or just changing it in `replace` is enough due to lack of other mechanism of enrofcement? (in that case #330 would be non issue

dkropachev:
> > Which means that when you upgrade `gocqlx` you need to upgrade `github.com/scylladb/gocql` version in the replace directive manually.
> 
> I see, I kinda forgot that in this require there is gocql/gocql no scylladb/gocql, can we somehow require scylladb version not the upstream one? 

unfortunately not.
We just need to migrate to scylladb/gocql at some point, it will fix all the problems
>Or just changing it in `replace` is enough due to lack of other mechanism of enforcement? (in that case #330 would be non issue

yes, unfortunately it is only way.

Issue #340, opened 2025-09-05
migrate fails if .cql file ends with a comment


kapitan-kid:
**Title:**  
Migration fails if `.cql` file ends with a comment  

**Description:**  
When `migrate.FromFS(...)` processes a `.cql` file that ends with a comment starting with `--`, the migration fails.  

**Steps to reproduce:**  
1. Create a `.cql` file with the following content:  
   ```
   CREATE TABLE IF NOT EXISTS some_table
   (
       some_column            ASCII,
       PRIMARY KEY ((some_column))
   ) WITH default_time_to_live = 3600; -- ttl 1 hour
   ```

2. Run the migration using migrate.FromFS(...).

**Actual result:**  
The migration fails with the following error:  
``Migration failed: apply migration "01_example.cql": statement 2: line 1:13 : Unexpected '' (potentially executed: false)
exit status 1``  

**Expected result:**
The migration should succeed. A trailing comment at the end of the file should not cause an error.

**Additional info:**
It looks like the issue occurs [here](https://github.com/scylladb/gocqlx/blob/master/migrate/migrate.go#L244). `stmt` is not empty, but it only contains a comment.

itsfuad:
Looking into it

dkropachev:
@itsfuad , thanks a lot for great PR, merged and release as part of https://github.com/scylladb/gocqlx/releases/tag/v3.0.4

kapitan-kid:
> [@kapitan-kid](https://github.com/kapitan-kid) Can you checkout my [codebase](https://github.com/itsfuad/gocqlx) and test if the solution worked? Then I'll create pull request to merge it.

Perfect, it works!  
I'm amazed at how quickly you fixed it.  Thank you very much.

PR #344, opened 2025-09-13
Fix: #340 Add comment handling and related tests for migration functionality

itsfuad:
This pull request improves the migration system by introducing logic to correctly handle SQL comments in migration files, ensuring that regular comments (including trailing comments after statements) are ignored during migration execution. It also adds comprehensive tests to verify this behavior and documents the new logic for clarity.

**Migration logic improvements:**

* Added the `isComment` function to distinguish regular SQL comments from callback commands, ensuring only valid statements and callbacks are processed during migrations (`migrate/migrate.go`).
* Updated the migration execution flow so that regular comments and empty statements are silently skipped, preventing errors from trailing comments in migration files (`migrate/migrate.go`).

**Testing enhancements:**

* Added the `TestMigrationWithTrailingComment` test to verify that migrations with trailing comments succeed without error (`migrate/migrate_test.go`).
* Added the `TestIsComment` test suite to validate the behavior of the new `isComment` function for various statement types (`migrate/migrate_test.go`).
* Exposed the `IsComment` function for use in tests and other packages (`migrate/export_test.go`).

**Documentation updates:**

* Documented the new migration logic and statement handling in the comments for the `applyMigration` function, clarifying the distinction between SQL statements, callback commands, and regular comments (`migrate/migrate.go`).

**What fixed**
https://github.com/scylladb/gocqlx/issues/340

### DIFF
--- a/migrate/export_test.go
+++ b/migrate/export_test.go
@@ -3,3 +3,7 @@ package migrate
 func IsCallback(stmt string) (name string) {
 	return isCallback(stmt)
 }
+
+func IsComment(stmt string) bool {
+	return isComment(stmt)
+}
--- a/migrate/migrate.go
+++ b/migrate/migrate.go
@@ -201,6 +201,23 @@ func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


dkropachev:
@itsfuad , thanks, what a great PR.

PR #376, opened 2026-07-28
qb: quote mixed-case table names

### LINKED ISSUE #168: Support for uppercase table name
https://github.com/scylladb/gocqlx/blob/master/qb/select.go#L83
https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/ucase-lcase_r.html

Unless enclosed in double quotation mark, identifier is case-insensitive. Currently, construrct like this:
```
table.New(table.Metadata{
  Name:    "aBc", 
  ...
})
```
doesn't work unless you manually add quotes to name, which is creepy and seems likely to break something in the future. 
Possibly, the same problem exists for some other identifiers. 

The solution seems to be to always enclose identifiers in double quotation mark in generated statements.

dkropachev:
## Changes
- Quote table name segments that need CQL quoting before rendering SELECT, INSERT, UPDATE, and DELETE statements.
- Preserve lowercase identifiers and already quoted table names for compatibility.
- Add coverage for keyspace-qualified mixed-case table names via table CRUD helpers.

Fixes #168

## Testing
- go test ./qb ./table
- go test $(go list ./... | rg -v '/cmd/schemagen$')
- go test ./... (fails: cmd/schemagen requires Scylla/Cassandra on 127.0.0.1:9042)

### DIFF
--- a/qb/doc.go
+++ b/qb/doc.go
@@ -4,4 +4,12 @@
 
 // Package qb provides CQL query builders. The builders create CQL statement
 // and a list of named parameters that can later be bound using gocqlx.
+//
+// Table names passed to statement builders may be keyspace-qualified. Unquoted
+// identifiers that require CQL quoting, such as mixed-case names or reserved
+// keywords, are quoted automatically while already quoted identifiers are
+// preserved. This intentionally changes generated CQL for mixed-case unquoted
+// names: Foobar is rendered as "Foobar" instead of Foobar, preserving case
+// rather than relying on CQL's lowercase folding. Pass lowercase names when
+// targeting lowercase tables.
 package qb
--- a/qb/utils.go
+++ b/qb/utils.go
@@ -35,6 +35,149 @@ func (cols columns) writeCql(cql *bytes.Buffer) {
 	}
 }
 
+func quoteTableName(table string) string {
+	if !strings.ContainsAny(table, `."`) {
+		return quoteIdentifier(table)
+	}
+
+	parts := splitTableName(table)
+	for i := range parts {
+		parts[i] = quoteIdentifier(parts[i])
+	}
+	return strings.Join(parts, ".")
+}
+
+func splitTableName(table string) []string {
+	var parts []string
+	start := 0
+	quoted := false
+
+	for i := 0; i < len(table); i++ {
+		switch table[i] {
+		case '"':
+			switch {
+			case quoted && i+1 < len(table) && table[i+1] == '"':
+				i++
+			case quoted:
+				quoted = false
+			case i == start:
+				quoted = true
+			}
+		case '.':
+			if !quoted {
+				parts = append(parts, table[start:i])
+				start = i + 1
+			}
+		}
+	}
+
+	return append(parts, table[start:])
+}
+
+func quoteIdentifier(identifier string) string {
+	if isQuotedIdentifier(identifier) || !needsQuoting(identifier) {
+		return identifier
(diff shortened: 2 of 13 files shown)


qb/utils.go
```diff
@@ -35,6 +35,72 @@ func (cols columns) writeCql(cql *bytes.Buffer) {
...
+				continue
+			}
+			inQuote = !inQuote
+		case '.':
+			if !inQuote {
+				parts = append(parts, name[start:i])
+				start = i + 1
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dkropachev: Fixed in a372c76: splitTableName now enters quoted mode only when the quote starts a segment, so quotes inside unquoted identifiers stay literal and get escaped by quoteIdentifier. Added regression coverage for key"space.tableName.

qb/utils.go
```diff
@@ -35,6 +35,131 @@ func (cols columns) writeCql(cql *bytes.Buffer) {
...
+func quoteIdentifier(identifier string) string {
+	if isQuotedIdentifier(identifier) || !needsQuoting(identifier) {
+		return identifier
+	}
+	return `"` + strings.ReplaceAll(identifier, `"`, `""`) + `"`
+}
+
+func isQuotedIdentifier(identifier string) bool {
+	return len(identifier) >= 2 && identifier[0] == '"' && identifier[len(identifier)-1] == '"'
+}
```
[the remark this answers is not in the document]
dkropachev: Fixed in e62a8f5: isQuotedIdentifier now preserves only quoted identifiers whose interior quotes are properly escaped as "". Invalid quoted segments are re-quoted with every quote escaped, and regression tests cover the injection-shaped inputs.

qb/doc.go
```diff
@@ -4,4 +4,9 @@
 
 // Package qb provides CQL query builders. The builders create CQL statement
 // and a list of named parameters that can later be bound using gocqlx.
+//
+// Table names passed to statement builders may be keyspace-qualified. Unquoted
+// identifiers that require CQL quoting, such as mixed-case names or reserved
+// keywords, are quoted automatically while already quoted identifiers are
+// preserved.
```
sylwiaszunejko: As I agree that this is the right behavior, I wonder if there are users out there that where some how relying on previous behavior and this would be a breaking change for them

qb/utils.go
```diff
@@ -35,6 +35,145 @@ func (cols columns) writeCql(cql *bytes.Buffer) {
 	}
 }
 
+func quoteTableName(table string) string {
+	parts := splitTableName(table)
```
sylwiaszunejko, line 43: I wonder about the performance impact of allocating `[]string` on every `ToCql` call, maybe there is a possibility to have a fast path for "no dot, no quote" case?

qb/doc.go
dkropachev: Fixed in e7eb504: the package docs now explicitly call out the behavior change for mixed-case unquoted names, including that Foobar now renders as "Foobar" and callers should pass lowercase names when targeting lowercase tables.

qb/utils.go
dkropachev: Fixed in e7eb504: quoteTableName now has a fast path for table names without dots or quotes, avoiding splitTableName and the []string allocation on the common simple-name path while still preserving reserved-keyword and mixed-case quoting.

sylwiaszunejko (approved):

Issue #375, opened 2026-07-24
schemagen: track tuple support for nested generated types


dkropachev:
## Problem

PR #374 improves schemagen's recursive CQL type parsing for nested frozen UDT collections, but full tuple support has to remain out of scope for now.

The blocker is in gocql tuple unmarshalling: scylladb/gocql#956. `gocql.Unmarshal` currently decodes tuple elements into generic temporary values from `TypeInfo.NewWithError()` and then assigns those values into the destination fields. That can panic for generated tuple structs when the generated field type differs from the temporary type.

Examples:

- `list<tuple<int>>` maps to `[]struct { Field1 int32 }`, but gocql decodes the tuple element as `int` before assigning to `int32`.
- `tuple<boolean, map<text, frozen<set<album>>>>` maps to a field like `map[string][]AlbumUserType`, but gocql can decode the element as `map[string][]map[string]any` before assignment.

## Scope for gocqlx

Until scylladb/gocql#956 is fixed and released into gocqlx, schemagen should not claim end-to-end support for nested tuple shapes that require concrete generated element types.

Follow-up work after the gocql fix:

- add schemagen/gocqlx round-trip tests for tuple fields containing primitive aliases, UDTs, and nested UDT collections
- update schemagen's tuple handling to support those shapes end to end, or explicitly reject any remaining unsupported tuple forms with a clear error
- update PR #374 documentation/scope notes once tuple support is safe to expand

PR #378, opened 2026-07-28
tuple: support arrays and slices

### LINKED ISSUE #152: Improve support for Tuples
We should be able to scan and bind arrays and slices.

At the moment

```go
func datatypesTuplesSimple(t *testing.T, session gocqlx.Session) {
	err := session.ExecStmt(`CREATE KEYSPACE IF NOT EXISTS examples WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}`)
	if err != nil {
		t.Fatal("create keyspace:", err)
	}

	err = session.ExecStmt(`CREATE TABLE IF NOT EXISTS examples.tuples(k int PRIMARY KEY, c tuple<int,int>)`)
	if err != nil {
		t.Fatal("create table:", err)
	}

	coordinates1 := [2]int{12, 34}
	coordinates2 := [2]int{56, 78}

	insert := session.Query(qb.Insert("examples.tuples").Columns("k").TupleColumn("c", 2).ToCql())
	insert.BindMap(qb.M{
		"k":   1,
		"c_0": coordinates1[0],
		"c_1": coordinates1[1],
	})
	if err := insert.Exec(); err != nil {
		t.Fatal("Exec() failed:", err)
	}
	insert.BindMap(qb.M{
		"k":   2,
		"c_0": coordinates2[0],
		"c_1": coordinates2[1],
	})
	if err := insert.Exec(); err != nil {
		t.Fatal("Exec() failed:", err)
	}

	row := &struct {
		K  int
		C [2]int
		C0 int `db:"c[0]"`
		C1 int `db:"c[1]"`
	}{}

	q := session.Query(qb.Select("examples.tuples").Columns("k", "c").ToCql())
	if err := q.Get(row); err != nil {
		t.Fatal(err)
	}
	t.Log(row)
}
```

At the moment it fails with `gocql: not enough columns to scan into: have 2 want 3`

dkropachev:
Closes #152

## Summary
- bind tuple placeholder names such as `c[0]` from non-byte array/slice and tuple-shaped struct fields, plus map values keyed by `c`
- expand tuple result columns into array, slice, or tuple-shaped struct fields, with explicit `db:"c[i]"` fields as fallback
- preserve tuple destinations when scans are exhausted or return no rows
- recognize tuple markers across supported CQL literals, comments, collection literals, and UDT literals
- cache tuple bind metadata per `Queryx` while retaining the low-allocation ordinary binding and scan paths
- reject byte sequences as tuple containers and preserve null/zero tuple elements through bind transformers

## Scope limitations
- Same-name tuple-shaped struct destinations are supported when they have one mapper-visible positional field per tuple element. Nested generated tuple type support remains tracked by #375 and scylladb/gocql#956.
- `[]byte` and byte arrays are scalar byte sequences, not tuple containers. Use `[][]byte` or an array/slice of non-byte elements for tuple values.
- Validation of non-positive tuple placeholder counts is pre-existing builder behavior and is intentionally deferred to #399.
- General `interface{}` destination support for ordinary, non-tuple struct scans is pre-existing behavior and is intentionally deferred to #400.

The deferred items are kept out of this PR to avoid expanding its scope beyond tuple array/slice support.

## Tests
- unit and race tests for the affected packages
- `go vet ./...`
- `go test -tags integration -run=NONE ./...`
- binding and scan-plan benchmarks

### DIFF
--- a/iterx.go
+++ b/iterx.go
@@ -23,9 +23,8 @@ type Iterx struct {
 	*gocql.Iter
 	Mapper *reflectx.Mapper
 
-	// Cache memory for a rows during iteration in structScan.
-	fields     [][]int
-	values     []interface{}
+	// Cache memory for rows during iteration in structScan.
+	scanPlan   *structScanPlan
 	strict     bool
 	structOnly bool
 	applied    bool
@@ -89,6 +88,10 @@ func (iter *Iterx) scanAny(dest interface{}) bool {
 			iter.err = structOnlyError(base)
 			return false
 		}
+	}
+
+	if tuple, ok := iter.topLevelTuple(base); ok {
+		return iter.scanTuple(value, tuple)
 	}
 
 	if scannable && len(iter.Columns()) > 1 {
@@ -156,6 +159,8 @@ func (iter *Iterx) scanAll(dest interface{}) bool {
 		}
 	}
 
+	tuple, tupleContainer := iter.topLevelTuple(base)
+
 	// if it's a base type make sure it only has 1 column;  if not return an error
 	if scannable && len(iter.Columns()) > 1 {
 		iter.err = fmt.Errorf("expected 1 column in result while scanning scannable type %s but got %d", base.Kind(), len(iter.Columns()))
@@ -173,9 +178,12 @@ func (iter *Iterx) scanAll(dest interface{}) bool {
 		vp = reflect.New(base)
 
 		// scan into the struct field pointers
-		if !scannable {
+		switch {
+		case tupleContainer:
+			ok = iter.scanTuple(vp, tuple)
+		case !scannable:
 			ok = iter.structScan(vp)
-		} else {
+		default:
 			ok = iter.scan(vp)
 		}
 		if !ok {
@@ -231,12 +239,66 @@ func (iter *Iterx) scan(value reflect.Value) bool {
 	return iter.Iter.Scan(udtWrapValue(value, iter.Mapper, iter.strict))
 }
 
+func (iter *Iterx) topLevelTuple(t reflect.Type) (gocql.TupleTypeInfo, bool) {
+	if !isTupleScanContainerType(t) {
+		return gocql.TupleTypeInfo{}, false
+	}
+
+	columns := iter.Columns()
+	if len(columns) != 1 {
+		return gocql.TupleTypeInfo{}, false
+	}
+
+	tuple, ok := columns[0].TypeInfo.(gocql.TupleTypeInfo)
+	return tuple, ok
+}
+
+func (iter *Iterx) scanTuple(value reflect.Value, tuple gocql.TupleTypeInfo) bool {
+	// Stage tuple allocation and unmarshalling so an exhausted iterator leaves
+	// the caller's destination unchanged. Keep using Iter.Scan so callers may
+	// safely alternate Scan and StructScan, including across page boundaries.
+	staged := reflect.New(value.Elem().Type())
+	staged.Elem().Set(value.Elem())
+	if staged.Elem().Kind() == reflect.Ptr {
+		if err := clonePointerChain(staged.Elem()); err != nil {
+			iter.err = err
+			return false
+		}
+	}
+
+	values := make([]interface{}, len(tuple.Elems))
+	for i := range tuple.Elems {
+		elem, err := tupleElementAddr(staged, i, len(tuple.Elems))
+		if err != nil {
+			iter.err = err
+			return false
+		}
+		if elem.Elem().Kind() == reflect.Interface {
+			values[i] = tupleInterfaceScanner{value: elem.Elem()}
+			continue
+		}
+		values[i] = udtWrapValue(elem, iter.Mapper, iter.strict)
+	}
+
+	if !iter.Iter.Scan(values...) {
+		return false
+	}
+	value.Elem().Set(staged.Elem())
+	return true
+}
+
 // StructScan is like gocql.Iter.Scan, but scans a single row into a single
 // struct. Use this and iterate manually when the memory load of Select() might
 // be prohibitive. StructScan caches the reflect work of matching up column
 // positions to fields to avoid that overhead per scan, which means it is not
 // safe to run StructScan on the same Iterx instance with different struct
 // types.
+//
+// Tuple columns can scan into non-byte array, slice, or tuple-shaped struct
+// fields with the same mapped name. Tuple-shaped structs honor mapper-visible
+// fields, including embedded fields and db:"-" exclusions. Element fields named
+// with db tags, such as `db:"coordinates[0]"`, remain supported. When both forms
+// are present, the field mapped to the tuple column takes precedence.
 func (iter *Iterx) StructScan(dest interface{}) bool {
 	value := reflect.ValueOf(dest)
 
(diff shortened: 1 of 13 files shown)


tuple.go
```diff
@@ -0,0 +1,193 @@
...
+		t = t.Field(index).Type
+		t = reflectx.Deref(t)
+	}
+	return t
+}
+
+func isTupleContainerType(t reflect.Type) bool {
+	t = reflectx.Deref(t)
+	return t.Kind() == reflect.Array || t.Kind() == reflect.Slice
+}
```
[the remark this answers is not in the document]
dkropachev: Fixed in 0eefd15. isTupleContainerType now returns false for nil before dereferencing. Root package tests pass.

iterx.go
```diff
@@ -286,12 +335,58 @@ func (iter *Iterx) structScan(value reflect.Value) bool {
...
+
+	for _, column := range columnInfo {
+		tuple, ok := column.TypeInfo.(gocql.TupleTypeInfo)
+		if !ok {
+			appendDestination(column.Name, traversalByName(iter.Mapper, t, column.Name), -1, 0, nil)
+			continue
+		}
+
+		traversal := traversalByName(iter.Mapper, t, column.Name)
+		if len(traversal) != 0 && isTupleContainerType(fieldTypeByTraversal(t, traversal)) {
```
nikagra: A tuple column whose mapped field is neither an array/slice nor per-element `db:"c[i]"` now goes entirely to discards — zero values, no error, and non-strict is the default. Before this PR it failed loudly with `not enough columns to scan into`. That hits our own codegen: `map_types.go` emits `struct{Field1 bool; Field2 int32}` for `tuple<boolean,int>`, so schemagen models silently lose tuple data. Either accept a struct with `NumField() == len(tuple.Elems)` here (gocql's `unmarshalTuple` already handles struct destinations) or error instead of discarding. Worth settling before merge.

iterx.go
```diff
@@ -237,6 +239,10 @@ func (iter *Iterx) scan(value reflect.Value) bool {
 // positions to fields to avoid that overhead per scan, which means it is not
 // safe to run StructScan on the same Iterx instance with different struct
 // types.
+//
+// Tuple columns can scan into array or slice fields with the same mapped name.
+// Element fields named with db tags, such as `db:"coordinates[0]"`, remain
+// supported.
```
nikagra: Precedence isn't stated: when a container field and `db:"c[0]"` fields both map to the same column, the container wins and the tagged fields stay zero — that's exactly the struct in #152. Worth spelling out here.

tuple.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	return t
+}
+
+func isTupleContainerType(t reflect.Type) bool {
+	if t == nil {
+		return false
+	}
+	t = reflectx.Deref(t)
+	return t.Kind() == reflect.Array || t.Kind() == reflect.Slice
```
nikagra: `[]byte` passes as a container, so `BindMap(qb.M{"c": []byte("ab")})` on a `tuple<int,int>` silently binds `97, 98`, while a `string` under the same key errors cleanly. `tuple<blob,blob>` needs `[][]byte`, so excluding `uint8` element types would give the same clear error.

iterx.go
```diff
@@ -24,11 +24,13 @@ type Iterx struct {
...
 	// Cache memory for a rows during iteration in structScan.
-	fields     [][]int
-	values     []interface{}
-	strict     bool
-	structOnly bool
-	applied    bool
+	fields       [][]int
+	tupleIndexes []int
+	tupleCounts  []int
+	values       []interface{}
```
nikagra: Four index-aligned slices plus `values` is easy to break, and `fieldsByTraversal` now takes five positional slices. Holding the `structScanPlan` itself would keep the invariant in one place and fix the cache guard as a side effect: `iter.fields == nil` never flips for a zero-column result, so the plan is rebuilt every row.

queryx.go
```diff
@@ -172,6 +181,7 @@ func (q *Queryx) SetHostID(hostID string) *Queryx {
 
 func (q *Queryx) bindStructArgs(arg0 interface{}, arg1 map[string]interface{}) ([]interface{}, error) {
 	arglist := make([]interface{}, 0, len(q.Names))
+	tupleCounts := tupleCountsByName(q.Names)
```
nikagra: `q.Names` is fixed for the life of a `Queryx`, but this re-walks it on every bind, and `tupleElementByName` does a fresh single-name `TraversalsByName` (plus a `[]string{name}` alloc) per element per bind. Computing it once in `Query()` keeps reused queries off that path; the nil map already keeps non-tuple queries allocation-free.

tuple.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	switch value.Kind() {
+	case reflect.Array, reflect.Slice:
+		if value.Len() != count {
+			return reflect.Value{}, fmt.Errorf("%s length %d does not match tuple element count %d", value.Kind(), value.Len(), count)
+		}
+		if index >= value.Len() {
+			return reflect.Value{}, fmt.Errorf("index %d out of range for %s of length %d", index, value.Kind(), value.Len())
+		}
```
nikagra: Two small things: an unset slice reports `slice length 0 does not match tuple element count 2`, which reads oddly for nil; and the `index >= value.Len()` check is unreachable — `count` is `max(index)+1` and the length is already pinned to `count`.

iterx_test.go
```diff
@@ -59,6 +59,64 @@ func diff(t *testing.T, expected, got interface{}) {
 
 var diffOpts = cmpopts.IgnoreUnexported(big.Int{}, inf.Dec{})
 
+func TestIterxTupleArraySlice(t *testing.T) {
```
nikagra, line 62: Only `tuple<int,int>` round-trips here. Null elements, `[]interface{}` (`tupleInterfaceScanner` is unit-tested only), tagged element fields, and `SetTuple` on UPDATE are all reflection-heavy and untested against a real cluster — worth one or two more cases.

iterx.go
dkropachev: Fixed in fab2ab3: a mapped same-name field with no tuple-element tags now returns a clear error instead of discarding. Full generated/nested struct support remains scoped to #375 and scylladb/gocql#956.

iterx.go
dkropachev: Fixed in fab2ab3: container-first scan precedence is now documented in StructScan and README, with a regression test using the #152 shape.

tuple.go
dkropachev: Fixed in fab2ab3: byte-element arrays/slices are rejected for binding and scanning; [][]byte remains valid for tuple<blob,...>. Added struct, map, and scan-plan tests.

iterx.go
dkropachev: Fixed in fab2ab3: Iterx now retains a single *structScanPlan, so initialization is explicit and the parallel-slice cache fields are removed.

queryx.go
dkropachev: Deferred to #397 and documented in the PR scope. The follow-up will benchmark this and account for public/mutable Names and Mapper, direct Queryx construction, and different bound struct types.

tuple.go
dkropachev: Keeping these as-is: a nil binding slice has length zero and does not provide the required tuple elements, while the bounds check is retained defensively at the helper boundary. No functional issue remains.

iterx_test.go
dkropachev: Added integration coverage in fab2ab3 for tagged element fields, null elements through []interface{}, and SetTuple on UPDATE.

iterx.go
dkropachev: Tuple-shaped struct fields now expand into element traversals, preserving schemagen data. Covered by unit and Scylla tests.

iterx.go
dkropachev: Documented tuple-column field precedence; added a mixed-field regression test.

tuple.go
dkropachev: Byte arrays and slices no longer count as tuple containers; nested byte-slice containers remain supported. Added bind and scan tests.

iterx.go
dkropachev: Iterx now caches structScanPlan directly, fixing the zero-column cache guard.

queryx.go
dkropachev: Deferred to #394 so caching semantics for public mutable Queryx.Names can be settled and benchmarked separately.

tuple.go
dkropachev: Nil slices now return a dedicated error; removed the unreachable bounds check.

iterx_test.go
dkropachev: Added Scylla coverage for tuple-shaped structs, tagged fields, nullable interface slices, and SetTuple updates.

tuple.go
```diff
@@ -0,0 +1,214 @@
...
+		return "", 0, false
+	}
+
+	open := strings.LastIndexByte(name, '[')
+	if open <= 0 {
+		return "", 0, false
+	}
+
+	index, err := strconv.Atoi(name[open+1 : len(name)-1])
+	if err != nil || index < 0 {
```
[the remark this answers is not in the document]
dkropachev: Fixed in 9a0933c. MaxInt tuple indexes are rejected before count calculation; added binding regression coverage.

queryx.go
dkropachev: Disposition finalized: this performance-only caching work is deferred to #394. The PR scope now references #394 consistently, and duplicate #397 has been closed. Resolving this thread as intentionally out of scope for the tuple correctness change.

queryx.go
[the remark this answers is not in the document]
dkropachev: Fixed in the subsequent revisions: every bindStructArgs error path returns both the argument slice and the error. The current branch compiles, passes lint, and passes the focused and race-enabled tests.

dkropachev:
Addressed the remaining review feedback in 6cd5aa8:

- restored byte-array/slice rejection to match the documented non-byte container contract; `[][]byte` remains supported
- made unsupported same-name struct bind fields return an explicit type error instead of falling back to `BindStructMap` values
- added coverage for tuple-shaped structs with unexported fields

The earlier body-only nits are also covered: README alignment, tuple-name parser edge cases, zero-column scan-plan caching, and the stale `fieldsByTraversal` comment. The license-header grammar suggestion was intentionally left unchanged because the same project-standard header is used across the repository.

Validation: focused tuple tests, `make check`, race-enabled non-database tests, and `go vet ./...` all pass.

queryx.go
sylwiaszunejko: `q.Names` is fixed for the life of the `Queryx`, so this is recomputed identically on every
single bind. For a query with tuple names it also allocates a fresh `map[string]int` each time.`q.Names` is fixed for the life of the `Queryx`, so this is recomputed identically on every
single bind. For a query with tuple names it also allocates a fresh `map[string]int` each time.

This can impact performance negatively, can we compute it at most once per query somehow?

queryx.go
```diff
@@ -172,6 +181,7 @@ func (q *Queryx) SetHostID(hostID string) *Queryx {
 
 func (q *Queryx) bindStructArgs(arg0 interface{}, arg1 map[string]interface{}) ([]interface{}, error) {
```
sylwiaszunejko, line 194: `README.md` states the scan rule: "If a struct also contains fields tagged with tuple element
names such as `db:"coordinate[0]"`, the array or slice field takes precedence when scanning."
`structScanPlan` implements that — it looks up the tuple column name first and only falls
through to element names when nothing maps to the column
([iterx.go:400-431](iterx.go#L400-L431)).

`bindStructArgs` does the **opposite**. `TraversalsByNameFunc` is asked for the full name
`"coordinates[0]"` first, and `tupleElementByName` is consulted only when that lookup misses
([queryx.go:192-213](queryx.go#L192-L213)). So the element-tagged fields win on the way out.

iterx.go
```diff
@@ -254,68 +299,180 @@ func (iter *Iterx) StructScan(dest interface{}) bool {
...
 		iter.err = err
 		return false
 	}
 
 	// scan into the struct field pointers and append to our results
-	return iter.Iter.Scan(iter.values...)
+	return iter.Iter.Scan(iter.scanPlan.values...)
+}
+
+func (iter *Iterx) structScanPlan(t reflect.Type, columnInfo []gocql.ColumnInfo) (structScanPlan, error) {
```
sylwiaszunejko: having this function named the same as the struct it is returning may be confusing

tuple.go
nikagra: Reopening: this isn't in the squashed head. `isTupleContainerType` at `99505a11` is still `t.Kind() == reflect.Array || t.Kind() == reflect.Slice`, and there's no `Uint8` check anywhere in the tree, so `BindMap(qb.M{"c": []byte("ab")})` still binds `97, 98`. The `queryx_test.go` "tuple byte slice" case now asserts that behaviour, and the PR description still says byte sequences are rejected — looks like `fab2ab3` was lost in the squash.

tuple.go
nikagra: Half of this is in the head: the dedicated nil-slice error is there, but the unreachable bounds checks are still at `tuple.go:123`, `156` and `160`.

tuple.go
```diff
@@ -0,0 +1,222 @@
...
+		return reflect.Value{}, err
+	}
+
+	switch value.Kind() {
+	case reflect.Array:
+		if value.Len() != count {
+			return reflect.Value{}, fmt.Errorf("array length %d does not match tuple element count %d", value.Len(), count)
+		}
+	case reflect.Slice:
+		if value.Len() != count {
```
nikagra: The slice is only reallocated when `Len() != count`, so a destination reused across rows keeps the same backing array. The manual `for iter.StructScan(&r)` loop that `StructScan`'s own doc comment recommends therefore ends with every retained copy holding the last row's tuple:

```go
var r struct{ K int; C []int }
for iter.StructScan(&r) { out = append(out, r) }
// rows (1,(1,2)),(2,(3,4)) -> [{1 [3 4]} {2 [3 4]}]
```

`Select` is safe because it does a fresh `reflect.New` per row. gocql's own list/set path always `MakeSlice`s, so this is the only slice-valued column that aliases.

queryx.go
```diff
@@ -183,10 +193,21 @@ func (q *Queryx) bindStructArgs(arg0 interface{}, arg1 map[string]interface{}) (
 		if len(t) != 0 {
 			val := reflectx.FieldByIndexesReadOnly(v, t)
 			arglist = append(arglist, val.Interface())
+		} else if val, ok, err := tupleElementByName(q.Mapper, v, q.Names[i], tupleCounts); err != nil {
```
nikagra: This runs before the `arg1` lookup, so `BindStructMap` can no longer bind an explicitly supplied `"c[0]"` map key when the struct has a same-named non-container field — `tupleElementByName` finds `C string`, fails `isTupleContainerType` and returns `expected array or slice in "c" but got string` without ever consulting the map. That worked before the PR (empty traversal fell through to `arg1`). Suggest trying the explicit `arg1` key before erroring on the struct field.

tuple.go
```diff
@@ -0,0 +1,222 @@
...
+	// The inferred tuple count is index + 1, so the largest int cannot be
+	// represented as a valid tuple element index.
+	if err != nil || index < 0 || index == math.MaxInt {
+		return "", 0, false
+	}
+
+	return name[:open], index, true
+}
+
+func tupleCountsByName(names []string) map[string]int {
```
nikagra: Arity is inferred from any bind name shaped `x[N]`, so an ordinary CQL collection-element assignment gets hijacked: `qb.Update(t).Set("l[0]")` emits the name `l[0]` verbatim, `counts["l"]` becomes 1, and a struct with the list field `L []int{7,8,9}` fails with `slice length 3 does not match tuple element count 1` — or, if `len(L) == 1`, silently binds `L[0]` instead of the intended value. `qb` already knows the real arity from `TupleColumn`/`SetTuple`; carrying it on the `Queryx` would avoid re-deriving it by string-parsing every name.

iterx.go
```diff
@@ -231,12 +239,49 @@ func (iter *Iterx) scan(value reflect.Value) bool {
 	return iter.Iter.Scan(udtWrapValue(value, iter.Mapper, iter.strict))
 }
 
+func (iter *Iterx) topLevelTuple(t reflect.Type) (gocql.TupleTypeInfo, bool) {
```
nikagra, line 242: `topLevelTuple`/`isTupleContainerType` test only the reflect kind, so a named array or slice that implements `gocql.Unmarshaler` is decomposed element-wise and its own `UnmarshalCQL` never runs — `scanAny` reaches this branch before the `scannable` path, and `StructOnly()` can't suppress it because it errors for non-struct kinds. Same on the bind side: `tupleElementByName` decomposes a slice type implementing `gocql.Marshaler`. Worth excluding types that implement the (un)marshaler interfaces, the way `isScannable` does.

queryx.go
```diff
@@ -216,11 +241,19 @@ func (q *Queryx) BindMap(arg map[string]interface{}) *Queryx {
...
 func (q *Queryx) bindMapArgs(arg map[string]interface{}) ([]interface{}, error) {
 	arglist := make([]interface{}, 0, len(q.Names))
+	tupleCounts := tupleCountsByName(q.Names)
 
 	for _, name := range q.Names {
 		val, ok := arg[name]
 		if !ok {
-			return arglist, fmt.Errorf("could not find name %q in %#v", name, arg)
+			var err error
+			val, ok, err = tupleElementFromMap(arg, name, tupleCounts)
```
nikagra: A nil tuple element comes back from here as an untyped nil and then flows into `q.tr`, where `UnsetEmptyTransformer` panics on `reflect.ValueOf(nil).IsZero()` — that's the `[]interface{}{nil, 90}` shape the new integration test inserts. Plain `BindMap(qb.M{"x": nil})` has always had this, so not a regression, but the tuple half is new: a legitimately zero element (e.g. `[2]float64{0, 19.94}`) becomes `gocql.UnsetValue` inside a `(?,?)` literal, which the server rejects.

iterx.go
```diff
@@ -254,68 +299,180 @@ func (iter *Iterx) StructScan(dest interface{}) bool {
...
+			switch {
+			case isTupleContainerType(fieldType):
+				for i := range tuple.Elems {
+					appendDestination(gocql.TupleColumnName(column.Name, i), traversal, i, len(tuple.Elems), nil)
+				}
+				continue
+			case fieldType != nil && fieldType.Kind() == reflect.Struct && fieldType.NumField() == len(tuple.Elems):
+				for i := range tuple.Elems {
+					field := fieldType.Field(i)
+					if field.PkgPath != "" {
```
nikagra: This walks the destination positionally with `NumField()`/`Field(i)`, ignoring the mapper, so db tags, `db:"-"` and embedding are all invisible. `C struct{ Base; Y int }` matches a 2-tuple and gives element 0 the traversal of the embedded `Base`, failing at scan with `can not unmarshal int into *Base`; conversely a struct with two int fields plus a `db:"-"` string field has `NumField() == 3` and is rejected outright even though the mapper sees exactly two usable fields. Driving this off `Mapper.TypeMap`/traversals would handle both.

tuple.go
```diff
@@ -0,0 +1,222 @@
...
+	traversal := traversalByName(mapper, value.Type(), base)
+	if len(traversal) == 0 {
+		return reflect.Value{}, false, nil
+	}
+
+	field, err := fieldByTraversalReadOnly(value, traversal)
+	if err != nil {
+		return reflect.Value{}, false, fmt.Errorf("could not bind tuple element %q: %w", name, err)
+	}
+	if !isTupleContainerType(field.Type()) {
```
nikagra: Binding doesn't accept the tuple-shaped structs that scanning now supports, so the same field is readable but not writable. `Get(&place)` fills `Coordinate Coord` via the new struct branch in `structScanPlan`, but `BindStruct(place)` fails here with `expected array or slice in "coordinate" but got Coord` — the struct can't round-trip through insert+select.

tuple.go
```diff
@@ -0,0 +1,222 @@
...
+	if !strings.HasSuffix(name, "]") {
+		return "", 0, false
+	}
+
+	open := strings.LastIndexByte(name, '[')
+	if open <= 0 {
+		return "", 0, false
+	}
+
+	index, err := strconv.Atoi(name[open+1 : len(name)-1])
```
nikagra: `strconv.Atoi` accepts non-canonical indexes like `c[+1]` and `c[007]`, but the scan side only ever emits `gocql.TupleColumnName` canonical form. So a field tagged `db:"c[+1]"` binds as element 1 while `structScanPlan` looks up only `c[1]` and hands the field a `tupleDiscardScanner`. `Names{"c[0]","c[+1]","c[1]"}` is worse: `counts["c"]` is 2 so the arity check passes, and element 1 is bound twice. Worth requiring a canonical decimal index.

iterx.go
```diff
@@ -254,68 +299,180 @@ func (iter *Iterx) StructScan(dest interface{}) bool {
...
+		if !ok {
+			appendDestination(column.Name, traversalByName(iter.Mapper, t, column.Name), -1, 0, nil)
+			continue
+		}
+
+		traversal := traversalByName(iter.Mapper, t, column.Name)
+		if len(traversal) != 0 {
+			fieldType := fieldTypeByTraversal(t, traversal)
+			switch {
+			case isTupleContainerType(fieldType):
```
nikagra: Array arity isn't validated at plan time here, unlike the struct branch below which checks `NumField`. `C [3]int` against a 2-tuple plans fine and then fails on every row with `array length 3 does not match tuple element count 2` — no column name, no type. `fieldType.Len()` is statically known at this point, so this could fail during planning with the column name.

qb/insert.go
```diff
@@ -142,6 +142,7 @@ func (b *InsertBuilder) FuncColumn(column string, fn *Func) *InsertBuilder {
 }
 
 // TupleColumn adds an insert column for a tuple value to the query.
+// It generates bind names in the form "column[0]", "column[1]", and so on.
 func (b *InsertBuilder) TupleColumn(column string, count int) *InsertBuilder {
```
nikagra, line 146: For `count <= 0` `tupleParam.writeCql` emits `(?)` plus the name `column[-1]`, which `parseTupleElementName` rejects, so the bind fails with `could not find name "c[-1]"` rather than anything about the count. Now that this comment documents the naming contract, `TupleColumn`/`SetTuple` could reject `count < 1` up front.

iterx.go
```diff
@@ -254,68 +299,180 @@ func (iter *Iterx) StructScan(dest interface{}) bool {
...
+				return err
+			}
+			if elem.Elem().Kind() == reflect.Interface {
+				plan.values[i] = tupleInterfaceScanner{value: elem.Elem()}
+				continue
+			}
+			plan.values[i] = udtWrapValue(elem, iter.Mapper, iter.strict)
+			continue
+		}
+		if plan.tupleCounts[i] > 0 && f.Kind() == reflect.Interface {
```
nikagra, line 559: The `interface{}` destination fix is gated on `tupleCounts[i] > 0`, so it only helps tuple elements — a plain `interface{}` field on a text column still goes through `f.Addr()` to `*interface{}` and fails in gocql. The same reflect-kind test also appears in `scanTuple` and twice in this function; folding it into `udtWrapValue` would remove all three special cases and fix non-tuple columns for free.

tuple.go
```diff
@@ -0,0 +1,222 @@
...
+		return reflect.Value{}, fmt.Errorf("tuple element index %d out of range for %s of length %d", index, value.Kind(), value.Len())
+	}
+	elem := value.Index(index)
+	if !elem.CanAddr() {
+		return reflect.Value{}, fmt.Errorf("cannot address tuple element %d in %s", index, value.Type())
+	}
+	return elem.Addr(), nil
+}
+
+func fieldByTraversalReadOnly(value reflect.Value, traversal []int) (reflect.Value, error) {
```
nikagra, line 548: `fieldByTraversalReadOnly` and `derefTupleValue` re-implement `reflectx.FieldByIndexesReadOnly` and `reflectx.FieldByIndexes`, which already deref and allocate nil pointers — `queryx.go:194` uses the real one two lines above the new copy, so the two walkers can drift on nil/embedded-pointer handling. Separately the `name[i]` format now lives in three places (`qb/value.go`, this parser, and `gocql.TupleColumnName` on the scan side); having `qb` emit `gocql.TupleColumnName` would stop producer and parser from diverging.

tuple.go
dkropachev: Fixed.

queryx.go
dkropachev: Fixed.

queryx.go
dkropachev: Fixed.

tuple.go
dkropachev: Fixed.

queryx.go
dkropachev: Fixed.

tuple.go
dkropachev: Fixed.

iterx.go
dkropachev: Fixed.

iterx.go
dkropachev: Fixed.

tuple.go
dkropachev: Fixed.

tuple.go
dkropachev: Fixed.

iterx.go
dkropachev: Fixed.

tuple.go
dkropachev: Fixed.

iterx.go
dkropachev: Renamed for clarity.

queryx.go
dkropachev: Fixed; tuple null and zero values are preserved.

qb/insert.go
dkropachev: Deferred to #399; this is pre-existing.

iterx.go
dkropachev: Deferred to #400; this is pre-existing.

tuple.go
dkropachev: Tuple name generation is centralized. The custom walkers remain for explicit nil and allocation handling.

queryx.go
```diff
@@ -216,21 +274,59 @@ func (q *Queryx) BindMap(arg map[string]interface{}) *Queryx {
...
+	if q.tr == nil {
+		return value
+	}
+
+	transformed := q.tr(name, value)
+	if tupleElement && reflect.TypeOf(transformed) == reflect.TypeOf(gocql.UnsetValue) {
+		return value
+	}
+	return transformed
+}
```
sylwiaszunejko, line 308: `transformBindValue` discards a transformer's `UnsetValue` for
any name the statement classifies as a tuple element. `tupleElement` is `tuple.count != 0`, which is decided by the **statement**, not by whether the
value was sourced from a container. So it also fires for the pre-existing `db:"c[0]"` element-field
style, which has nothing to do with this feature. Anyone combining UnsetEmptyTransformer with qb.TupleColumn/SetTuple/*Tuple comparators silently starts overwriting tuple columns on upgrade.

The change in behavior should be fixed or documented properly.

queryx.go
```diff
@@ -172,28 +193,60 @@ func (q *Queryx) SetHostID(hostID string) *Queryx {
...
+		case tuple.count != 0:
+			element, found, tupleErr := tupleElementByName(q.Mapper, v, name, tuple)
+			switch {
+			case tupleErr == nil && found:
+				val = element.Interface()
+			case tupleErr != nil:
+				// An explicit BindStructMap element remains a valid fallback when
+				// the struct has a same-name field that is not a tuple container.
+				var ok bool
+				if val, ok = arg1[name]; !ok {
```
sylwiaszunejko, line 219: `bindStructArgs` checks `tupleErr != nil` before `len(directTraversal) != 0`, so a struct with both a same-name non-container field and its own element fields now errors. Hit by e.g. `qb.InTuple("id", 2)`. `arg1["id[0]"]` is honoured; the struct's own field isn't. Fix: try `directTraversal` before returning `tupleErr`.

queryx.go
```diff
@@ -216,21 +274,59 @@ func (q *Queryx) BindMap(arg map[string]interface{}) *Queryx {
 
 func (q *Queryx) bindMapArgs(arg map[string]interface{}) ([]interface{}, error) {
 	arglist := make([]interface{}, 0, len(q.Names))
+	tupleElements := q.cachedTupleBindElements()
 
-	for _, name := range q.Names {
+	for i, name := range q.Names {
 		val, ok := arg[name]
```
sylwiaszunejko, line 280: Container vs `c[0]` element, same input: `BindStruct → [10 20]` (container), `BindMap → [99 98]` (element keys), `StructScan` → container. Not a regression, but public contract once released.

PR #379, opened 2026-07-28
queryx: add unsafe scan mode

### LINKED ISSUE #365: Allow `Queryx.Unsafe()`
Right now, if I want to leave out some fields from be scanned into a `struct`, I have do to something like
```go
query.Iter().Unsafe().Select(&myStruct);
...
defer query.Release()
```
This took me quite some time searching through the code to find out, and the documentation for `QueryX.Iter()` doesn't even suggest this is a valid use case.

Ideally, I could invoke instead:
```go
query.Unsafe().SelectRelease(&myStruct)
```
This would be similar to the `DB.Unsafe()` method and easier to use.

If I'm not missing something, I guess I could create a PR for it.

dkropachev:
## Summary
- add `Unsafe()` to `Queryx` so callers can opt out of strict missing-field checks before `Get`, `Select`, or `SelectRelease`
- add `Unsafe()` to `Iterx` for the equivalent iterator-level opt-out
- cover strict-to-unsafe query and iterator flows in the existing strict scan tests

Fixes #365

## Testing
- `go test $(go list ./... | rg -v '/cmd/schemagen$')`\n- `go test ./...` fails locally because `cmd/schemagen` expects Scylla/Cassandra on `127.0.0.1:9042`

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,35 @@ fmt.Println(people)
 // stdout: [{Michał Matczuk [michal@scylladb.com]}]
 ```
 
+## Strict struct scanning
+
+By default, GocqlX ignores result columns that do not map to exported fields in
+the destination struct. Use `Strict` to catch schema or query drift when scanning
+structs:
+
+```go
+var p Person
+q := session.Query(`SELECT first_name, last_name, email, age FROM person`, nil).Strict()
+if err := q.GetRelease(&p); err != nil {
+	log.Fatal(err) // age has no matching Person field
+}
+```
+
+Set `gocqlx.DefaultStrict = true` to make new queries strict by default. Use
+`Unsafe` on a query or iterator to ignore unmapped columns for that query or
+iterator instance:
+
+```go
+var p Person
+q := session.Query(`SELECT * FROM person`, nil).Unsafe()
+if err := q.GetRelease(&p); err != nil {
+	log.Fatal(err)
+}
+```
+
+When binding UDT values with `Queryx.Bind`, call `Unsafe` before `Bind` so the
+UDT wrapper is created in unsafe mode.
+
 ## Generating table metadata with schemagen
 
 Installation
--- a/iterx.go
+++ b/iterx.go
@@ -13,8 +13,11 @@ import (
 	"github.com/scylladb/go-reflectx"
 )
 
-// DefaultStrict disables the behavior of forcing queries and iterators to ignore
-// missing fields for all queries. See Strict below for more information.
+// DefaultStrict makes newly created queries use strict struct and UDT mapping.
+// In strict mode, struct scans report an error when a result column has no
+// matching destination field, and UDT marshal/unmarshal reports an error when a
+// UDT field name has no matching struct field. Use Queryx.Unsafe or
+// Iterx.Unsafe to disable strict mode for an individual query or iterator.
 var DefaultStrict bool
 
 // Iterx is a wrapper around gocql.Iter which adds struct scanning capabilities.
@@ -31,11 +34,22 @@ type Iterx struct {
 	applied    bool
 }
 
-// Strict forces the iterator to disable ignoring missing fields. In Strict mode
-// when scanning a struct if result row has a column that cannot be mapped to any
-// destination field an error is reported. By default such columns are ignored.
+// Strict makes the iterator use strict struct and UDT mapping. Struct scans
+// report an error when a result column cannot be mapped to any destination
+// field, and UDT marshal/unmarshal reports an error when a UDT field name has
+// no matching struct field. By default unmatched names are ignored unless
+// DefaultStrict is enabled. Use Unsafe to disable strict mode again for this
+// iterator.
 func (iter *Iterx) Strict() *Iterx {
 	iter.strict = true
+	return iter
+}
+
+// Unsafe makes the iterator ignore result columns and UDT fields that cannot be
+// mapped to destination struct fields. It can override DefaultStrict or a prior
+// Strict call; the last Strict or Unsafe call on the iterator wins.
+func (iter *Iterx) Unsafe() *Iterx {
+	iter.strict = false
 	return iter
 }
 
--- a/iterx_test.go
+++ b/iterx_test.go
@@ -760,6 +760,16 @@ func TestIterxStrict(t *testing.T) {
 		golden = "missing destination name \"testtextunbound\" in gocqlx_test.StrictTable"
 	)
 
+	setDefaultStrict := func(t *testing.T, strict bool) {
+		t.Helper()
+
+		defaultStrict := gocqlx.DefaultStrict
+		gocqlx.DefaultStrict = strict
+		t.Cleanup(func() {
+			gocqlx.DefaultStrict = defaultStrict
+		})
+	}
+
 	t.Run("get strict", func(t *testing.T) {
 		var v StrictTable
 		err := session.Query(stmt, nil).Strict().Get(&v)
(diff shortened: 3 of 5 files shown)


sylwiaszunejko:
First, I was a bit confused by this PR, to be honest. Are we sure the user who reported this is using v3 of `gocqlx`, where Strict mode was introduced? They didn't mention enabling Strict mode and then wanting to opt out because they couldn't. This use case seems pretty rare to me. 
If this is a real use case, maybe we could call this added function `DisableStrict` or something similar to be consistent in terms of naming?

iterx.go
```diff
@@ -31,14 +34,25 @@ type Iterx struct {
...
+// DefaultStrict is enabled. Use Unsafe to disable strict mode again for this
+// iterator.
 func (iter *Iterx) Strict() *Iterx {
 	iter.strict = true
 	return iter
 }
 
+// Unsafe makes the iterator ignore result columns and UDT fields that cannot be
+// mapped to destination struct fields. It can override DefaultStrict or a prior
+// Strict call; the last Strict or Unsafe call on the iterator wins.
```
nikagra, line 50: Two claims here don't hold:

1. *"the last Strict or Unsafe call on the iterator wins"* — only before the first scan. `iter.fields` and the missing-column check are latched once, on the first `structScan` (iterx.go:276-292), whereas UDT wrapping re-reads `iter.strict` on every row (iterx.go:319). A call made mid-iteration therefore flips UDT mapping but leaves column checking on its original setting.
2. *"can override DefaultStrict"* — `Iterx` never reads `DefaultStrict`. It inherits `strict` from `Queryx.Iter()` (queryx.go:377); a bare `&Iterx{}` is non-strict regardless of the global, which is exactly what the new `TestIterxStrictUnsafe` relies on.

Suggested wording:

```go
// Unsafe makes the iterator ignore result columns and UDT fields that cannot be
// mapped to destination struct fields, overriding the mode inherited from the
// query. Call it before the first scan: the missing-column check is evaluated
// once, when the first row is scanned.
```

The "before the first scan" caveat applies equally to `Strict` above (pre-existing), so it may be worth stating on both.

queryx.go
```diff
@@ -378,10 +378,22 @@ func (q *Queryx) Iter() *Iterx {
...
 
-// Strict forces the query and iterators to report an error if there are missing fields.
-// By default when scanning a struct if result row has a column that cannot be mapped to
-// any destination it is ignored. With strict error is reported.
+// Strict makes the query and iterators created from it use strict struct and UDT
+// mapping. Struct scans report an error when a result column cannot be mapped
+// to any destination field. UDT marshal/unmarshal reports an error when a UDT
+// field name has no matching struct field. By default these unmatched names are
+// ignored unless DefaultStrict is enabled. Use Unsafe to disable strict mode
+// again for this query.
```
nikagra, line 386: `Unsafe` below correctly warns to call it before `Bind`, because `Bind` captures `q.strict` into the UDT wrapper at bind time (queryx.go:237 → `udtWrapSlice`). `Strict` carries exactly the same constraint and doesn't mention it, so the two freshly rewritten comments now disagree about whether ordering matters.

Cleanest fix: state the rule once on `Bind` — the strict/unsafe mode in effect when `Bind` is called is the one used for UDT values — and have both `Strict` and `Unsafe` point at it.

iterx_test.go
```diff
@@ -779,6 +789,100 @@ func TestIterxStrict(t *testing.T) {
...
+		if err != nil {
+			t.Fatal("Select() failed:", err)
+		}
+		if len(v) != 1 {
+			t.Fatalf("Select()=%+v expected 1 row got %d", v, len(v))
+		}
+		if diff := cmp.Diff(m, v[0]); diff != "" {
+			t.Fatalf("Select()[0]=%+v expected %+v, diff: %s", v[0], m, diff)
+		}
+	})
```
nikagra, line 884: The struct-scan half of `Unsafe` is well covered here, but the UDT half isn't covered at all: `grep "missing name" *_test.go` returns nothing and there's no `udt_test.go`, so the strict branches in `udt.MarshalUDT`/`udt.UnmarshalUDT` (udt.go:45,56) have never had a test. That matters for this PR specifically, because it promotes UDT strictness to documented API surface *and* adds a non-obvious ordering rule about it (`Unsafe` before `Bind`).

Concretely: reverting `Bind` to v2's behavior — `udtWrapSlice(q.Mapper, DefaultUnsafe, v)`, ignoring the per-query flag — would still pass CI as written.

Suggested addition, reusing the existing `FullNameUDT` fixture in this file: with `setDefaultStrict(t, true)`, bind a UDT value whose Go type is missing one of the UDT's fields and assert the `missing name %q` error; then the same bind via `.Unsafe().Bind(...)` and assert success.

iterx_test.go
```diff
@@ -760,6 +760,16 @@ func TestIterxStrict(t *testing.T) {
...
 
+	setDefaultStrict := func(t *testing.T, strict bool) {
+		t.Helper()
+
+		defaultStrict := gocqlx.DefaultStrict
+		gocqlx.DefaultStrict = strict
+		t.Cleanup(func() {
+			gocqlx.DefaultStrict = defaultStrict
+		})
+	}
```
nikagra, line 771: Nit: `setDefaultStrict` is a closure local to `TestIterxStrict`, while queryx_test.go:216-219 duplicates the same save/restore inline — one shared helper would do.

Also worth a one-line comment that it must not be combined with `t.Parallel()`. Mutating the `DefaultStrict` global is only safe here because these tests are serial, and `batchx_test.go` in this same package already uses `t.Parallel()`.

README.md
```diff
@@ -147,6 +147,35 @@ fmt.Println(people)
...
+```go
+var p Person
+q := session.Query(`SELECT * FROM person`, nil).Unsafe()
+if err := q.GetRelease(&p); err != nil {
+	log.Fatal(err)
+}
+```
+
+When binding UDT values with `Queryx.Bind`, call `Unsafe` before `Bind` so the
+UDT wrapper is created in unsafe mode.
```
nikagra, line 177: Nit / adjacent gap, pre-existing and not this PR's job: `Batch.BindStruct`/`BindMap`/`Bind` (batchx.go:63,75,84,95) pass args straight to `b.Query(...)` without `udtWrapSlice`, so `Strict()`/`Unsafe()` on a `Queryx` handed to a batch has no effect on UDT values at all. Since this PR is what documents UDT strictness, either note the limitation next to this sentence or open a follow-up issue.

nikagra (commented):
Implementation looks correct — `Queryx.Iter()` already copies `strict` (queryx.go:377), so the `Unsafe().Get/Select/SelectRelease` flow from #365 works, and restoring the v2 `Unsafe()` name (v2.8.0/iterx.go:38) is a nice migration affordance. Mutating the receiver rather than returning a copy differs from sqlx's `DB.Unsafe()`, but it matches the existing `Strict()`/`StructOnly()` style here, so I'd keep it. The `DefaultStrict` doc rewrite is a real improvement over the old double negative.

Verified locally at 870cc11: `go test . ./qb/... ./table/...` → 41 pass; `go vet .` and `go vet -tags integration .` clean; both new unit tests pass. No performance or security impact, and there's no CHANGELOG in this repo to update.

Two things I'd want before merge: the `Iterx.Unsafe` doc comment makes two claims that don't hold (inline), and the UDT half of strict/unsafe — which this PR is what documents — has no test at all. Two non-blocking nits also inline.

sylwiaszunejko:
I don't think there is a need for this PR any more, the original issue is closed, because the user wasn't using gocqlx v3 and that was the root cause of the issue

dkropachev:
Closing since the reported use case was on v2; v3 already ignores unmapped fields by default. We can revisit a per-query `DefaultStrict` override separately if a concrete need arises.

PR #381, opened 2026-07-28
examples: add batch and update flows

### LINKED ISSUE #256: Provide example for batches and updates
A couple of examples would be helpful.

`qb` package appears deprecated, but there is no other apparent support for batches in currently latest version (2.8.0) either.

As there is no evident way to extract values from `gocqlx.Queryx` for `gocql.Batch.Query`  function either, it is not clear how `gocql.Batch` is supposed to be used with gocqlx now.

Similarly with updates, there is only Insert/Select examples, but an update example would be also appreciated.

dkropachev:
## Summary
- add an example that binds multiple structs into a gocqlx batch and verifies the complete inserted rows
- add an example that updates a row through a table model
- factor the shared songs/playlists example schema into reusable helpers
- correct the playlist table model clustering key to match the schema, preventing Get, Update, and Delete queries from filtering on the regular `song_id` column

Fixes #256

## Testing
- `go test $(go list ./... | rg -v "/cmd/schemagen$")`
- `make start-scylla && go test -tags all -count=1 -run "^TestExample$" .; make stop-scylla`

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/gocql/gocql"
+	"github.com/google/go-cmp/cmp"
 	"golang.org/x/sync/errgroup"
 	"gopkg.in/inf.v0"
 
@@ -24,7 +25,7 @@ import (
 )
 
 // Running examples locally:
-// make run-scylla
+// make start-scylla
 // make run-examples
 func TestExample(t *testing.T) {
 	cluster := gocqlxtest.CreateCluster()
@@ -38,6 +39,8 @@ func TestExample(t *testing.T) {
 	_ = session.ExecStmt(`DROP KEYSPACE examples`)
 
 	basicCreateAndPopulateKeyspace(t, session, "examples")
+	basicBatchInsertStructs(t, session, "examples")
+	basicUpdateWithTableModel(t, session, "examples")
 	createAndPopulateKeyspaceAllTypes(t, session)
 	basicReadScyllaVersion(t, session)
 
@@ -77,10 +80,7 @@ type PlaylistItem struct {
 	SongID gocql.UUID
 }
 
-// This example shows how to use query builders and table models to build
-// queries. It uses "BindStruct" function for parameter binding and "Select"
-// function for loading data to a slice.
-func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspace string) {
+func createMusicTables(t *testing.T, session gocqlx.Session, keyspace string) {
 	t.Helper()
 
 	if err := gocqlxtest.CreateKeyspaceIfNotExists(session, keyspace); err != nil {
@@ -108,14 +108,33 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
 	if err != nil {
 		t.Fatal("create table:", err)
 	}
+}
 
-	playlistMetadata := table.Metadata{
+func songTableModel(keyspace string) *table.Table {
+	return table.New(table.Metadata{
+		Name:    fmt.Sprintf("%s.songs", keyspace),
+		Columns: []string{"id", "title", "album", "artist", "tags", "data"},
+		PartKey: []string{"id"},
+	})
+}
+
+func playlistTableModel(keyspace string) *table.Table {
+	return table.New(table.Metadata{
 		Name:    fmt.Sprintf("%s.playlists", keyspace),
 		Columns: []string{"id", "title", "album", "artist", "song_id"},
 		PartKey: []string{"id"},
-		SortKey: []string{"title", "album", "artist", "song_id"},
-	}
-	playlistTable := table.New(playlistMetadata)
+		SortKey: []string{"title", "album", "artist"},
+	})
+}
+
+// This example shows how to use query builders and table models to build
+// queries. It uses "BindStruct" function for parameter binding and "Select"
+// function for loading data to a slice.
+func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspace string) {
+	t.Helper()
+
+	createMusicTables(t, session, keyspace)
+	playlistTable := playlistTableModel(keyspace)
 
 	// Insert song using query builder.
 	insertSong := qb.Insert(fmt.Sprintf("%s.songs", keyspace)).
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -108,14 +107,33 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+
+func playlistTableModel(keyspace string) *table.Table {
+	return table.New(table.Metadata{
 		Name:    fmt.Sprintf("%s.playlists", keyspace),
 		Columns: []string{"id", "title", "album", "artist", "song_id"},
 		PartKey: []string{"id"},
-		SortKey: []string{"title", "album", "artist", "song_id"},
-	}
-	playlistTable := table.New(playlistMetadata)
+		SortKey: []string{"title", "album", "artist"},
```
nikagra, line 126: This is a real fix, not a cosmetic tweak, and it's worth calling out.

The table is declared `PRIMARY KEY (id, title, album, artist)`, so the old `SortKey` listed `song_id` — a regular column. That was harmless while `playlistTable` was only used for `Insert` (columns) and `Select` (partition key only), but `Table.Get` builds its `WHERE` from the full primary key. Without this change the new `GetQuery` below would emit `... AND song_id = ?` and Scylla would reject it with the *ALLOW FILTERING* error.

Could you mention it in the PR body, and ideally split it into its own commit? It fixes a latent defect in a README-linked example that people copy, so it deserves to be findable rather than buried in a refactor diff.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
 	}
 }
 
+// This example shows how to add structs to a batch and execute them together.
+func batchInsertStructs(t *testing.T, session gocqlx.Session) {
```
nikagra: nit: the other example helpers in this file carry a doc-section prefix — `basic*`, `datatypes*`, `paging*`, `lwt*`. Consider `basicBatchInsertStructs`, or introduce an explicit `batch` section, to keep the file navigable.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+	t.Helper()
+
+	createMusicTables(t, session, "examples")
```
nikagra: nit: every sibling helper takes a `keyspace` parameter, and that shape is load-bearing — `batchx_test.go` calls `basicCreateAndPopulateKeyspace(t, session, "batch_test")` precisely because of it. Hardcoding `"examples"` here (6 occurrences across the two new functions) means these examples can't be reused from a real test the same way.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+	}
+	playlist := PlaylistItem{
+		ID:     mustParseUUID("7720de96-8984-41a7-a560-91f2398d6ebf"),
+		Title:  song.Title,
+		Album:  song.Album,
+		Artist: song.Artist,
+		SongID: song.ID,
+	}
+
+	insertSong := songTableModel("examples").InsertQuery(session)
```
nikagra: nit: `songTableModel("examples")` is called again on line 224, and `playlistTableModel("examples")` on line 209 and again on 232 — four `table.New` calls, each re-running the qb builders for `get`/`sel`/`insert`.

Table models are meant to be built once and reused (typically package-level). `updateWithTableModel` gets this right with a single local; since this file is documentation people copy from, it'd be good for the batch example to model the same thing. Two locals would do it.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+	defer insertSong.Release()
```
nikagra: nit: this is correct — `Batch.BindStruct` builds a fresh arg slice (`bindStructArgs` allocates per call) and stores only `qry.Statement()`, so releasing before `ExecuteBatch` can't alias pooled state.

It does slightly obscure that the `Queryx` here is only a statement+names holder, though. Releasing right after binding might read more clearly in an example.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+	insertPlaylist := playlistTableModel("examples").InsertQuery(session)
+	defer insertPlaylist.Release()
+
+	batch := session.Batch(gocql.LoggedBatch)
```
nikagra, line 212: nit: worth one line on *why* `LoggedBatch`. Keeping the denormalized `songs` + `playlists` pair in sync is the textbook legitimate multi-partition logged batch, so this is the right choice here — but readers copy examples verbatim, and single-partition batches are the cheap default. A short comment stops this becoming a cargo-culted multi-partition batch.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+	}
+	if err := session.ExecuteBatch(batch); err != nil {
+		t.Fatal("execute batch:", err)
+	}
+
+	var gotSong Song
+	if err := songTableModel("examples").GetQuery(session).BindStruct(song).GetRelease(&gotSong); err != nil {
+		t.Fatal("select song:", err)
+	}
+	if gotSong.Title != song.Title {
```
nikagra: nit: this asserts only `Title`, and the playlist check below only `SongID`. Enough to prove both statements in the batch applied, but `TestBatch` in `batchx_test.go` uses `cmp.Diff(got, want)` over the whole struct — stronger, and consistent with the existing style.

(`Tags` is a `set<text>`, so ordering comes back server-normalized — fine for a single element, just worth knowing if you compare whole structs.)

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+	if err := playlistTableModel("examples").GetQuery(session).BindStruct(playlist).GetRelease(&gotPlaylist); err != nil {
+		t.Fatal("select playlist:", err)
+	}
+	if gotPlaylist.SongID != song.ID {
+		t.Fatalf("playlist song_id=%v expected %v", gotPlaylist.SongID, song.ID)
+	}
+}
+
+// This example shows how to update a row through a table model.
+func updateWithTableModel(t *testing.T, session gocqlx.Session) {
```
nikagra: nit: same naming point as `batchInsertStructs` — no doc-section prefix, unlike the rest of the file.

example_test.go
```diff
@@ -164,6 +182,95 @@ func basicCreateAndPopulateKeyspace(t *testing.T, session gocqlx.Session, keyspa
...
+		Tags:   []string{"demo"},
+		Data:   []byte("music"),
+	}
+	if err := songTable.InsertQuery(session).BindStruct(song).ExecRelease(); err != nil {
+		t.Fatal("insert song:", err)
+	}
+
+	song.Title = "Updated title"
+	song.Album = "Updated album"
+	if err := songTable.UpdateQuery(session, "title", "album").BindStruct(song).ExecRelease(); err != nil {
```
nikagra, line 273: nit: worth a comment noting that the full `song` struct is bound here but only `title` and `album` are written. That binding is driven by the statement's named parameters — not by what's in the struct — is exactly the thing users get wrong, and it's really the point of this example.

nikagra (approved):
LGTM. This answers #256 directly — `Batch.BindStruct(qry, arg)` is exactly the missing link the issue reporter couldn't find — and the flows are genuinely exercised, not just compiled: CI's `make test` runs `go test -tags all .`, which executes `TestExample` against a live Scylla container, and all checks pass. The refactor into `createMusicTables` / `songTableModel` / `playlistTableModel` is a good seam; `batchx_test.go` already depends on the parameterized shape of `basicCreateAndPopulateKeyspace`, so factoring in that direction fits.

The inline comments are all non-blocking nits, with one exception worth reading: the `SortKey` change is a real latent-bug fix and currently reads like a cosmetic tweak.

Two notes on the PR description itself:

- The Testing section has a literal `\n` in it, so both commands render on one line.
- `go test -tags integration -run '^$' ./...` compiles the package but runs nothing. `make run-scylla && make run-examples` is what actually exercises these examples (`-run=Example` does match `TestExample`). Worth listing that instead, since the `SortKey` change and the batch flow both need a live cluster to validate.

example_test.go
dkropachev: Called this out explicitly in the PR body and retained the corrected `SortKey`. The branch is kept as a single squashed commit as requested.

example_test.go
dkropachev: Renamed the helper to `basicBatchInsertStructs` to match the basic-section naming convention.

example_test.go
dkropachev: Added a `keyspace` parameter to both new helpers and passed `examples` from `TestExample`; setup and model construction now use the parameter.

example_test.go
dkropachev: Created one `songTable` and one `playlistTable` local and reused each for its insert and get queries.

example_test.go
dkropachev: Moved each `Release` immediately after `Batch.BindStruct`. The bind error is captured first, so the query is released on both success and error paths before batch execution.

example_test.go
dkropachev: Added a comment explaining that a logged batch is used because the denormalized rows span partitions and must stay in sync.

example_test.go
dkropachev: Changed both checks to whole-struct `cmp.Diff` comparisons, matching `batchx_test.go`.

example_test.go
dkropachev: Renamed the helper to `basicUpdateWithTableModel` to match the basic-section naming convention.

example_test.go
dkropachev: Added a comment explaining that the full struct can be bound because the statement names select `title`, `album`, and primary-key `id`.

PR #382, opened 2026-07-28
examples: add basic runnable app

### LINKED ISSUE #319: Add example for goclqx
Have an example app that leverages gocqlx.
Example should include `go.mod`

dkropachev:
## Summary
- add `examples/basic`, a standalone runnable gocqlx app with its own `go.mod`
- demonstrate creating schema, inserting with a table model, and reading rows with `SelectRelease`
- link the app from the root README examples section

Fixes #319

## Testing
- `go test ./...` from `examples/basic`
- `go test $(go list ./... | rg -v '/cmd/schemagen$')` from the repository root


examples/basic/main.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, part := range parts {
+		if host := strings.TrimSpace(part); host != "" {
+			out = append(out, host)
+		}
+	}
+	return out
+}
+
+func createSchema(session gocqlx.Session, keyspace string) error {
+	if err := session.ExecStmt(fmt.Sprintf(`CREATE KEYSPACE IF NOT EXISTS %s WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}`, keyspace)); err != nil {
```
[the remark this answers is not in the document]
dkropachev: Fixed in c7a8f8e. Switched the example keyspace to NetworkTopologyStrategy with replication_factor 1 and removed the SimpleStrategy/tablet-disable retry. Verified with make check, example unit tests, and a smoke run against current Scylla.

examples/basic/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+	session, err := gocqlx.WrapSession(cluster.CreateSession())
+	if err != nil {
+		return fmt.Errorf("create session: %w", err)
+	}
+	defer session.Close()
+
+	if err := createSchema(session, *keyspace); err != nil {
+		return fmt.Errorf("create schema: %w", err)
+	}
```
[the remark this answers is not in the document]
dkropachev: Addressed in 503783c: the example now calls `AwaitSchemaAgreement` with a bounded 10-second context after schema creation and before the insert, and returns the error with context.

examples/basic/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+	people := table.New(table.Metadata{
+		Name:    *keyspace + ".people",
+		Columns: []string{"id", "first_name", "last_name"},
+		PartKey: []string{"id"},
```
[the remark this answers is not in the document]
dkropachev: Addressed in 503783c: keyspace input is restricted to the CQL keyspace-name limits, quoted before interpolation, and the same quoted qualified name is used by both schema DDL and table-model queries. Tests cover invalid input, mixed case, DDL, and generated INSERT CQL.

examples/basic/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+func createSchema(session execStmtSession, keyspace string) error {
+	stmt := fmt.Sprintf(`CREATE KEYSPACE IF NOT EXISTS %s WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}`, keyspace)
+	if err := session.ExecStmt(stmt); err != nil {
+		return err
+	}
+	return session.ExecStmt(fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s.people (
+		id uuid PRIMARY KEY,
+		first_name text,
+		last_name text)`, keyspace))
```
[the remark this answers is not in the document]
dkropachev: Keeping `NetworkTopologyStrategy` here. Cassandra added `replication_factor` auto-expansion for NTS in 4.0 (CASSANDRA-14303), and the Cassandra 4.0 DDL reference documents this exact syntax: https://cassandra.apache.org/doc/4.0/cassandra/cql/ddl.html#networktopologystrategy. Scylla documents the same behavior and recommends NTS over SimpleStrategy: https://docs.scylladb.com/manual/stable/cql/ddl.html#networktopologystrategy. The example now explicitly scopes Cassandra support to 4.0+, while retaining compatibility with tablet-enabled Scylla.

examples/basic/README.md
```diff
@@ -0,0 +1,11 @@
+# Basic gocqlx Example
+
+This example creates a keyspace and table, inserts a row with a table model, and reads it back with `SelectRelease`.
+
+Run a local Scylla node, then run:
+
+```bash
+go run . -hosts 127.0.0.1:9042
+```
```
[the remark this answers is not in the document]
dkropachev: Addressed in 503783c: the README now says the commands start at the repository root and includes `cd examples/basic` before `go run .`.

Makefile
```diff
@@ -38,15 +38,18 @@ endif
...
 check: .require-golangci-lint
 	@golangci-lint run ./...
+	@cd ./examples/basic && golangci-lint run ./...
 
 .PHONY: fix
 fix: .require-golangci-lint .require-fieldalignment
 	@$(MAKE) fmt
 	@golangci-lint run --fix ./...
+	@cd ./examples/basic && golangci-lint run --fix ./...
 	@fieldalignment -test=false -fix  ./...
```
sylwiaszunejko, line 53: `fieldalignment` pass doesn't cover `examples/basic`, maybe it doesn't need to, but for consistency maybe worth adding

sylwiaszunejko (commented):
Looks good!

examples/basic/main_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if err := createSchema(session, `"example"`); err != nil {
+		t.Fatal(err)
+	}
+
+	want := []string{
+		`CREATE KEYSPACE IF NOT EXISTS "example" WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}`,
+		`CREATE TABLE IF NOT EXISTS "example".people (
+		id uuid PRIMARY KEY,
+		first_name text,
+		last_name text)`,
```
sylwiaszunejko, line 20: isn't hard-coding whitespaces a potential issue? shouldn't it be normalized somehow?

examples/basic/go.mod
```diff
@@ -0,0 +1,20 @@
...
+
+require (
+	github.com/google/uuid v1.6.0 // indirect
+	github.com/klauspost/compress v1.18.6 // indirect
+	github.com/scylladb/go-reflectx v1.0.1 // indirect
+	golang.org/x/sync v0.22.0 // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+)
+
+replace github.com/gocql/gocql => github.com/scylladb/gocql v1.18.3
```
nikagra: **Dependency pins drift from the rest of the repo.** Every other module here matches root exactly; this one doesn't, and the committed `go.sum` locks it in.

|  | scylladb/gocql | x/sync | klauspost/compress |
|---|---|---|---|
| root `go.mod` | v1.18.0 | v0.20.0 | v1.18.5 |
| `cmd/schemagen/testdata/go.mod` | v1.18.0 | v0.20.0 | — |
| **`examples/basic/go.mod`** | **v1.18.3** | **v0.22.0** | **v1.18.6** |

(verified with `go list -m` in each module)

Two concrete consequences:

- `make test`'s new smoke run exercises gocqlx against a driver version the library's own suite never sees, so a failure there can't be attributed to gocqlx vs. a driver delta without bisecting by hand.
- Renovate is `config:recommended`, so it auto-discovers Go modules. It will now manage three modules with nothing linking them, and at `prConcurrentLimit: 2` the spread widens one bump at a time.

Note also that `.gitignore` gitignores `cmd/schemagen/testdata/go.sum`, so that sibling module re-resolves on every `go mod tidy`, whereas this one is pinned by a committed `go.sum`. Committing `go.sum` is the better practice — it just means the drift has to be corrected by hand.

I applied the alignment locally and confirmed it's clean: `go mod tidy`, `go build ./...`, `go vet ./...` and `go test ./...` all pass on root's versions.

```suggestion
	github.com/google/uuid v1.6.0 // indirect
	github.com/klauspost/compress v1.18.5 // indirect
	github.com/scylladb/go-reflectx v1.0.1 // indirect
	golang.org/x/sync v0.20.0 // indirect
	gopkg.in/inf.v0 v0.9.1 // indirect
)

replace github.com/gocql/gocql => github.com/scylladb/gocql v1.18.0
```

⚠️ Committing that suggestion on its own leaves `go.sum` stale and **will break the build** — it needs 6 replacement hash lines. Run `go mod tidy` in `examples/basic` immediately after and commit the refreshed `go.sum` alongside it.

Longer term a root `go.work` would stop three hand-maintained copies drifting at all, but that's a bigger change than this PR should carry.

Makefile
nikagra, line 53: Heads-up that this line can't actually run today — and that's not this PR's fault.

`fix` lists `.require-fieldalignment` as a prerequisite (line 49), but no such target is defined anywhere in the Makefile and `get-tools` never installs the binary:

```
$ make -n fix
make: *** No rule to make target '.require-fieldalignment', needed by 'fix'.  Stop.
```

That's present on `master` too, so it predates this PR. It does mean the `examples/basic` `--fix` plumbing added here has never executed and can't until the prerequisite is either defined or dropped. Probably worth a separate issue rather than expanding the scope of #382.

The good news: `make check` genuinely does cover `examples/basic`, and since the shared `.golangci.yml` sets `govet: enable-all: true`, `fieldalignment` is enforced there through govet regardless of the broken `fix` target. So no lint coverage is lost — only the auto-fix convenience.

examples/basic/main.go
```diff
@@ -0,0 +1,135 @@
...
+	if len(rows) != 1 {
+		return fmt.Errorf("select person: got %d rows, want 1", len(rows))
+	}
+	if rows[0] != person {
+		return fmt.Errorf("select person: got %+v, want %+v", rows[0], person)
+	}
+	row := rows[0]
+	if _, err := fmt.Fprintf(os.Stdout, "%s %s %s\n", row.ID, row.FirstName, row.LastName); err != nil {
+		return fmt.Errorf("print person: %w", err)
+	}
```
nikagra: For an example whose whole purpose is to be copy-pasted, `fmt.Fprintf(os.Stdout, ...)` wrapped in an error check teaches a heavier idiom than the situation needs — three lines and an error path to print three fields. In library code that rigour is right; here `fmt.Printf` reads better:

```go
//nolint:forbidigo // example output
fmt.Printf("%s %s %s\n", row.ID, row.FirstName, row.LastName)
```

The `nolint` carries an explanation and names a specific linter, so it satisfies `nolintlint`'s `require-explanation` and `require-specific` settings.

I'd guess `forbidigo` is why it's written this way — it's enabled in the root config and its default pattern blocks `fmt.Print*`. I couldn't confirm that: I don't have `golangci-lint` locally, and CI reporting zero issues doesn't distinguish "root config picked up from the nested module and the code avoids forbidigo" from "config not discovered, defaults applied, forbidigo never ran". Either way the readability point stands, and if `forbidigo` turns out not to be active here the `nolint` can simply be dropped.

nikagra (commented):
Round 2. Re-reviewed at `503783c`. I ran the example locally: `go vet ./...` clean, `go test -count=1 ./...` 9/9 passing, plus a built binary probed against the flag-validation and empty-host paths. CI on head is green (Build, Snyk, CodeRabbit). I could not run `golangci-lint` locally, so lint rests on CI's `make check`.

### Prior findings — all five addressed

| Finding | Status |
|---|---|
| `SimpleStrategy` breaks on tablet-enabled Scylla | ✅ `NetworkTopologyStrategy` + RF 1, tablet-disable retry no longer needed |
| NTS needs per-DC settings on Cassandra | ✅ Correctly resolved by *documenting* Cassandra 4.0+ — the bare `replication_factor` shorthand for NTS landed in CASSANDRA-14303, so the README and flag help are accurate |
| Wait for schema agreement before insert | ✅ `MaxWaitSchemaAgreement` + bounded-ctx `AwaitSchemaAgreement` (`main.go:55-59`) |
| Validate/quote keyspace before CQL interpolation | ✅ Allowlist + double-quoting applied at **both** interpolation sites (DDL and `table.Metadata.Name`), with a test for the injection case |
| README must state the working directory | ✅ `cd examples/basic` added |

The keyspace fix is the strongest of these — it closes both interpolation sites rather than just the one that was flagged, `TestPeopleTableUsesQuotedKeyspace` pins the metadata path, and 48 is the correct Cassandra/Scylla identifier limit.

### Round 2

One should-fix (dependency pin drift) and two smaller notes, all inline. Optional nits:

- **Assertions in demo code** (`main.go:76-81`): `len(rows) != 1` and `rows[0] != person` read as a test rather than an example. It's doing double duty as the `make test` smoke check, which is defensible — but that role could move to an `integration`-tagged test, leaving `main.go` purely illustrative.
- **Case-sensitivity is undocumented** (`examples/basic/README.md:12`): because `quoteKeyspace` always double-quotes, `-keyspace Team` creates `"Team"`, which `cqlsh -e 'USE Team'` will not find. One sentence would save someone real confusion.
- **Root README coherence** (`README.md:230`): the new line sits directly above "Go and run them locally: `make run-scylla` / `make run-examples`", but `run-examples` is `go test -run=Example ./...` from the root and does not reach `examples/basic` — and `run-scylla` doesn't exist at all (the target is `start-scylla`; pre-existing). A reader following the next two lines won't run the new app. The sibling link also uses an absolute GitHub URL while this one is relative, so it won't resolve on pkg.go.dev.
- **`splitHosts`** (`main.go:89-98`): the `out := parts[:0]` in-place filter is correct — writes always trail reads — but the aliasing trick is gratuitous cleverness in example code; a plain `var out []string` reads better. Empty or whitespace-only `-hosts` yields `gocql: ... no hosts provided`, which I checked is a clear enough message.
- **Missing ScyllaDB licence header** on both new `.go` files. Soft convention only — 9 existing files also lack it (including `cmd/schemagen/testdata/models_test.go`), and `goheader` is enabled with no template configured, so it's currently a no-op.
- **`Person` field order** (`main.go:20-24`) is lint-driven (govet `fieldalignment`) rather than reading `ID` first. Harmless, but worth knowing in a file whose purpose is to be read.

CodeRabbit's "Docstring Coverage 0.00%" pre-merge warning isn't a lint failure — `revive`'s `package-comments` rule is disabled and this is `package main`. Still, doc comments on `quoteKeyspace` and `createSchema` would earn their keep here.

### Merge readiness

The code is correct, the three unit tests cover exactly the surfaces the first round attacked and need no database, and the Makefile/CI plumbing is thorough. Align the `examples/basic` dependency pins and I'd consider this ready — everything else above is optional polish.

Process note: `mergeable_state` is `blocked` while all checks are green. The existing review on this PR was submitted as a comment rather than an approval, so the required-approval gate still needs a formal approving review.

Makefile
dkropachev: it should not, it only matters for the core code

examples/basic/main_test.go
dkropachev: Fixed in ec575db: schema assertions now normalize whitespace while still checking statement count, order, and content.

Makefile
dkropachev: No change: the standalone example is covered by its own golangci-lint run, including govet/fieldalignment; the explicit fieldalignment fixer remains scoped to core code.

examples/basic/main.go
dkropachev: Fixed in ec575db: simplified the example output to fmt.Printf with a scoped, explained forbidigo suppression.

examples/basic/go.mod
dkropachev: Fixed in ec575db: aligned gocql, x/sync, and compress with the root module and regenerated go.sum.

Makefile
dkropachev: Agreed; no change here because the missing helper target predates this PR and is outside its scope. make check still lints the example.

sylwiaszunejko (approved):

PR #383, opened 2026-07-28
migrate: preserve progress after cancellation

### LINKED ISSUE #237: migrate: detach context from parent context cancelation
Migrate uses Query with Context [1](https://github.com/scylladb/gocqlx/blob/fc92258512d39bff365cae10d452a63cc1afc932/migrate/migrate.go#L243) [2](https://github.com/scylladb/gocqlx/blob/fc92258512d39bff365cae10d452a63cc1afc932/migrate/migrate.go#L190).

The problem is that the context may be canceled when migration statement is sent but before it's recorded in gocqlx migration table.

To solve this I suggest we do use [xcontext to detach](https://pkg.go.dev/golang.org/x/tools/internal/xcontext) from the context lifecycle and manually check if the context has been canceled before processing next statement.

dkropachev:
## Summary
- detach migration bookkeeping writes from parent context cancellation with `context.WithoutCancel`
- check the parent context before starting each next unapplied migration statement
- add an integration regression test that cancels after a callback and verifies recorded progress is preserved

Fixes #237

## Testing
- `go test ./migrate`
- `go test -tags integration -run '^$' ./migrate`\n- `go test $(go list ./... | rg -v '/cmd/schemagen$')`\n\nNot run locally: Scylla-backed execution of `TestMigrationRecordsProgressAfterContextCanceled`, because no Scylla/Cassandra node is running on `127.0.0.1:9042`.

### DIFF
--- a/migrate/callback.go
+++ b/migrate/callback.go
@@ -23,9 +23,15 @@ const (
 
 // CallbackFunc enables execution of arbitrary Go code during migration.
 // If error is returned the migration is aborted.
-// BeforeMigration and AfterMigration are triggered before and after processing
-// of each migration file respectively.
+// BeforeMigration is triggered before processing each migration file and again
+// before processing its remaining statements after a partial migration resumes.
+// AfterMigration is attempted after the final statement's progress is recorded
+// and is not retried if it returns an error, including one caused by context
+// cancellation or deadline expiry.
 // CallComment is triggered for each comment in a form `-- CALL <name>;` (note the semicolon).
+// BeforeMigration and AfterMigration receive the caller's context. CallComment
+// receives a context detached from caller cancellation and deadlines so a callback
+// that is part of a started migration operation completes with its progress update.
 type CallbackFunc func(ctx context.Context, session gocqlx.Session, ev CallbackEvent, name string) error
 
 // Callback is means of executing Go code during migrations.
--- a/migrate/migrate.go
+++ b/migrate/migrate.go
@@ -143,13 +143,18 @@ func Migrate(ctx context.Context, session gocqlx.Session, dir string) error {
 // FromFS executes new CQL files from a file system abstraction (io/fs.FS).
 // The provided FS has to be a flat directory containing *.cql files.
 //
+// Cancellation and deadlines stop migration between statements. Once a
+// statement starts, its execution and progress update finish before the
+// cancellation error is returned. AfterMigration runs only when the file's
+// final statement completes.
+//
 // It supports code based migrations, see Callback and CallbackFunc.
 // Any comment in form `-- CALL <name>;` will trigger an CallComment callback.
 func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
 	// get database migrations
 	dbm, err := List(ctx, session)
 	if err != nil {
-		return fmt.Errorf("list migrations: %s", err)
+		return fmt.Errorf("list migrations: %w", err)
 	}
 
 	// get file migrations
@@ -184,18 +189,18 @@ func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
 	if len(dbm) > 0 {
 		last := len(dbm) - 1
 		if err := applyMigration(ctx, session, f, fm[last], dbm[last].Done); err != nil {
-			return fmt.Errorf("apply migration %q: %s", fm[last], err)
+			return fmt.Errorf("apply migration %q: %w", fm[last], err)
 		}
 	}
 
 	for i := len(dbm); i < len(fm); i++ {
 		if err := applyMigration(ctx, session, f, fm[i], 0); err != nil {
-			return fmt.Errorf("apply migration %q: %s", fm[i], err)
+			return fmt.Errorf("apply migration %q: %w", fm[i], err)
 		}
 	}
 
 	if err = session.AwaitSchemaAgreement(ctx); err != nil {
-		return fmt.Errorf("awaiting schema agreement: %s", err)
+		return fmt.Errorf("awaiting schema agreement: %w", err)
 	}
 
 	return nil
@@ -211,7 +216,8 @@ func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
 // allowing for resumption of partially completed migrations.
 //
 // Parameters:
-//   - ctx: context for cancellation and timeouts
+//   - ctx: context checked between statements and passed to BeforeMigration,
+//     AfterMigration, and schema-agreement waits
 //   - session: database session for executing statements
 //   - f: filesystem containing the migration file
 //   - path: path to the migration file within the filesystem
@@ -244,12 +250,15 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 		"end_time",
 	).ToCql()
 
-	update := session.ContextQuery(ctx, stmt, names)
+	// Once a statement starts, allow both the statement and its progress update
+	// to finish. The parent context is checked between statements below.
+	operationCtx := context.WithoutCancel(ctx)
+	update := session.ContextQuery(operationCtx, stmt, names)
 	defer update.Release()
 
 	if DefaultAwaitSchemaAgreement.ShouldAwait(AwaitSchemaAgreementBeforeEachFile) {
 		if err = session.AwaitSchemaAgreement(ctx); err != nil {
-			return fmt.Errorf("awaiting schema agreement: %s", err)
+			return fmt.Errorf("awaiting schema agreement: %w", err)
 		}
 	}
 
@@ -274,15 +283,25 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 			continue
 		}
 
-		if Callback != nil && i == 1 {
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("context ended before statement %d: %w", i, err)
+		}
+
+		if Callback != nil && i == done+1 {
 			if err := Callback(ctx, session, BeforeMigration, info.Name); err != nil {
-				return fmt.Errorf("before migration callback: %s", err)
+				return fmt.Errorf("before migration callback: %w", err)
+			}
+			if err := ctx.Err(); err != nil {
+				return fmt.Errorf("context ended before statement %d: %w", i, err)
 			}
 		}
 
 		if DefaultAwaitSchemaAgreement.ShouldAwait(AwaitSchemaAgreementBeforeEachStatement) {
 			if err = session.AwaitSchemaAgreement(ctx); err != nil {
-				return fmt.Errorf("awaiting schema agreement: %s", err)
+				return fmt.Errorf("awaiting schema agreement before statement %d: %w", i, err)
+			}
+			if err := ctx.Err(); err != nil {
+				return fmt.Errorf("context ended before statement %d: %w", i, err)
 			}
 		}
 
@@ -295,14 +314,14 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 			if Callback == nil {
 				return fmt.Errorf("statement %d: missing callback handler while trying to call %s", i, cb)
 			}
-			if err := Callback(ctx, session, CallComment, cb); err != nil {
-				return fmt.Errorf("callback %s: %s", cb, err)
+			if err := Callback(operationCtx, session, CallComment, cb); err != nil {
+				return fmt.Errorf("callback %s: %w", cb, err)
 			}
 		} else if stmt != "" && !isComment(stmt) {
 			// Execute SQL statements (skip empty statements and comments)
-			q := session.ContextQuery(ctx, stmt, nil).RetryPolicy(nil)
+			q := session.ContextQuery(operationCtx, stmt, nil).RetryPolicy(nil)
 			if err := q.ExecRelease(); err != nil {
-				return fmt.Errorf("statement %d: %s", i, err)
+				return fmt.Errorf("statement %d: %w", i, err)
 			}
 		}
 		// Regular comments and empty statements are silently skipped
@@ -311,7 +330,7 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 		info.Done = i
 		info.EndTime = time.Now()
 		if err := update.BindStruct(info).Exec(); err != nil {
-			return fmt.Errorf("migration statement %d: %s", i, err)
+			return fmt.Errorf("migration statement %d: %w", i, err)
 		}
 	}
 	if i == 0 {
@@ -320,7 +339,7 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 
 	if Callback != nil && i > done {
 		if err := Callback(ctx, session, AfterMigration, info.Name); err != nil {
-			return fmt.Errorf("after migration callback: %s", err)
+			return fmt.Errorf("after migration callback: %w", err)
 		}
 	}
 
--- a/migrate/migrate_test.go
+++ b/migrate/migrate_test.go
@@ -9,11 +9,15 @@ package migrate_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"strings"
+	"sync"
 	"testing"
+	"time"
 
+	"github.com/gocql/gocql"
 	"github.com/psanford/memfs"
 
 	"github.com/scylladb/gocqlx/v3"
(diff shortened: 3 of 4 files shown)


migrate/migrate_test.go
```diff
@@ -379,6 +379,38 @@ func TestMigrationCallback(t *testing.T) {
 	})
 }
 
+func TestMigrationRecordsProgressAfterContextCanceled(t *testing.T) {
+	session := gocqlxtest.CreateSession(t)
+	defer session.Close()
+	recreateTables(t, session)
+
+	f := memfs.New()
+	err := f.WriteFile("0.cql", []byte(fmt.Sprintf(insertMigrate, 0)+";\n-- CALL cancel;\n"+fmt.Sprintf(insertMigrate, 1)+";"), fs.ModePerm)
```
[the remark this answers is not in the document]
dkropachev: Addressed: added a post-cancellation callback sentinel and assertions that it is skipped until the migration resumes.

migrate/migrate.go
sylwiaszunejko, line 274: shouldn't we check ctx.Err before break here? Or after the whole loop to not miss anything?

migrate/migrate.go
```diff
@@ -244,7 +244,11 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 		"end_time",
 	).ToCql()
 
-	update := session.ContextQuery(ctx, stmt, names)
+	// Once a migration operation starts, allow both the operation and its
+	// progress update to finish. The parent context is checked between
+	// statements below.
+	operationCtx := context.WithoutCancel(ctx)
```
sylwiaszunejko, line 255: `context.WithoutCancel` strips all done-channel signaling from the parent,
including a `context.WithTimeout`/`WithDeadline` set by the caller — not just
an explicit `cancel()` call, is this acceptable behavior here?

migrate/migrate.go
sylwiaszunejko: doc comment not true anymore

nikagra (commented):
Reviewed the fix end to end and traced both new tests against the statement parser and the `done`-resume logic. The core change is correct and matches what #237 prescribed — detach the bookkeeping write from the parent lifecycle, then check cancellation manually between statements. `context.WithoutCancel` is the right stdlib replacement for the issue's `xcontext` suggestion, and `go.mod` is at `go 1.25.0` so it's available.

I've skipped the two points @sylwiaszunejko already raised (deadline stripping at `:250`, stale docstring at `:214`) and replied in those threads instead.

### Verified

Traced both tests statement by statement; the expected values (`done` 2 then 4, counts 1 then 2, `afterCancelCalls` 0 then 1) are all correct.

The earlier concern that the regression test didn't actually exercise the new guard is genuinely resolved by the `after_cancel` sentinel. With the guard removed, statement 3 is `-- CALL after_cancel;` — a callback, so it runs on the *detached* context instead of failing, giving `done=3` and `afterCancelCalls=1`, and both assertions fail. `TestMigrationFinishesStartedStatementAfterContextCanceled` discriminates too: without the guard nothing errors at all, so it trips the `err == nil` check. Both tests now fail if the guard is deleted, which is what you want.

### Open items

Four inline comments. Only the callback-context one is a design question; the other three are mechanical.

### Docs

Beyond `applyMigration`'s docstring that @sylwiaszunejko flagged, `FromFS`'s doc makes the same implicit "cancellation and timeouts" promise, and `CallbackFunc` in `migrate/callback.go` says nothing about ctx semantics at all. That last one is the public contract and isn't in this diff, so it's the easiest to overlook.

### Tests

Both new tests are gated behind `//go:build all || integration`, and the PR description states the Scylla-backed run of `TestMigrationRecordsProgressAfterContextCanceled` didn't happen locally. So the entire test contribution is currently unexecuted — worth confirming CI actually runs these rather than only compiling them, since `-run '^$'` proves compilation only.

Coverage gaps I'd like to see closed:

- No `context.WithTimeout` case. Since `WithoutCancel` drops deadlines as well as cancellation, that's the untested half of the behavior change.
- No multi-file cancellation — cancel during file 2 of 3 and assert the state of files 1 and 3.
- No cancellation during the `AwaitSchemaAgreementBeforeEachFile` path, which still uses the parent ctx.

### Nit

`TestMigrationFinishesStartedStatementAfterContextCanceled` mutates the `DefaultAwaitSchemaAgreement` global and restores it with `defer`. That matches the surrounding style, so take it or leave it, but `t.Cleanup` also survives a `t.Fatal` inside a helper.

migrate/migrate.go
```diff
@@ -274,14 +278,18 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 			continue
 		}
 
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
 		if Callback != nil && i == 1 {
-			if err := Callback(ctx, session, BeforeMigration, info.Name); err != nil {
+			if err := Callback(operationCtx, session, BeforeMigration, info.Name); err != nil {
```
nikagra: Detaching `CallComment` at line 306 is clearly right: its completion increments `done`, so the callback and the bookkeeping write have to be atomic or you reintroduce exactly the bug #237 describes.

`BeforeMigration` here and `AfterMigration` at line 330 are a different case, though — neither is recorded in `gocqlx_migrate`, so detaching them buys no progress preservation. What it does buy is that arbitrary user code can no longer be interrupted, and there's no backstop for it: gocql bounds its own work via `ClusterConfig.Timeout` and `MaxWaitSchemaAgreement`, but nothing bounds a callback. A user whose `BeforeMigration` waits on something slow now has no way to abort the migration at all.

Two ways out, either fine as long as it's deliberate:

- pass the parent `ctx` to `BeforeMigration`/`AfterMigration` and keep `operationCtx` for `CallComment`; or
- keep the detach and document it on `CallbackFunc` in `migrate/callback.go` — that's the public contract, it currently says nothing about ctx semantics, and it isn't in this diff so users have no way to notice the change.

Worth flagging that the decision is currently being locked in by test: `migrate_test.go:399` asserts the callback's ctx is *not* canceled, so whichever way you go, that assertion encodes it.

migrate/migrate.go
```diff
@@ -274,14 +278,18 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 			continue
 		}
 
+		if err := ctx.Err(); err != nil {
+			return err
```
nikagra: This returns a bare `context.Canceled` with no indication of how far the migration got, which is the first thing an operator will want to know.

```suggestion
			return fmt.Errorf("canceled before statement %d: %w", i, err)
```

See the note on the test file about `%w` vs `%s` — `FromFS` currently flattens this with `%s`, so the `%w` here only pays off if that changes too.

migrate/migrate.go
```diff
@@ -274,14 +278,18 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
...
 		if Callback != nil && i == 1 {
-			if err := Callback(ctx, session, BeforeMigration, info.Name); err != nil {
+			if err := Callback(operationCtx, session, BeforeMigration, info.Name); err != nil {
 				return fmt.Errorf("before migration callback: %s", err)
 			}
 		}
 
 		if DefaultAwaitSchemaAgreement.ShouldAwait(AwaitSchemaAgreementBeforeEachStatement) {
-			if err = session.AwaitSchemaAgreement(ctx); err != nil {
+			if err = session.AwaitSchemaAgreement(operationCtx); err != nil {
```
nikagra: This wait uses `operationCtx`, but the before-each-file wait at line 255 still uses the parent `ctx`. I don't think there's a principled reason for the asymmetry.

Nothing is in flight for statement `i` at this point — we're immediately after the `ctx.Err()` guard at line 281 and before the statement executes — so aborting here loses no progress; everything through `i-1` is already recorded. Parent `ctx` would also be meaningfully more responsive, since `MaxWaitSchemaAgreement` defaults to 60s and this wait is currently unabortable.

Suggest parent `ctx` here to match line 255, or a comment explaining why the two differ.

migrate/migrate_test.go
```diff
@@ -379,6 +379,118 @@ func TestMigrationCallback(t *testing.T) {
...
+			afterCancelCalls++
+		}
+		return nil
+	}
+	defer func() {
+		migrate.Callback = nil
+	}()
+
+	err = migrate.FromFS(ctx, session, f)
+	if err == nil || !strings.Contains(err.Error(), context.Canceled.Error()) {
```
nikagra: This string match is worth pausing on: it's necessary because `FromFS` wraps with `%s` rather than `%w` —

https://github.com/scylladb/gocqlx/blob/d1e9fd2687f629387e938c6812d14f8625d40d45/migrate/migrate.go#L187

so the error chain is flattened and `errors.Is(err, context.Canceled)` returns false for callers.

That matters more after this PR than before it. Cancellation is now a routine, expected outcome, and a caller will want to distinguish it from "the migration is broken" — the first means "safe to re-run and resume", the second doesn't. Right now the only way to tell them apart is `strings.Contains` on the message, which is exactly what this test had to fall back on.

Suggest switching lines 187 and 193 to `%w`, then asserting `errors.Is(err, context.Canceled)` here and at line 472.

migrate/migrate.go
nikagra: Answering "is this acceptable here" — I think it depends on the call site, and it splits fairly cleanly.

**For the bookkeeping write and statement execution: acceptable.** Losing the caller's deadline there is essentially the point of the fix, and gocql still bounds the work independently of the context — `ClusterConfig.Timeout` per query, `MaxWaitSchemaAgreement` (60s default) for agreement waits. So a caller's `WithTimeout` stops applying, but execution stays bounded.

**For the callbacks: this is where it actually bites**, because user code has no equivalent backstop. A callback that blocks makes the migration uninterruptible with no upper bound at all. I left a longer note on the `BeforeMigration` call about splitting those cases from `CallComment`.

Either way, agreed on your other comment — and `FromFS`'s doc should stop implying timeout support too, not just `applyMigration`'s.

migrate/migrate.go
nikagra: I don't think a check is needed before the `break`, for two reasons:

- At EOF, every statement that was applied has already been recorded — the `update` write happens at the end of each loop iteration — so there's no progress left to lose here.
- Cancellation isn't silently swallowed either: after the file loop, `FromFS` calls `session.AwaitSchemaAgreement(ctx)` on the *parent* context, which fails on a canceled ctx. So `FromFS` still returns an error rather than reporting a clean success.

There is a real gap where your instinct is pointing, though it's after the loop rather than before the break: `AfterMigration` at line 330 runs on `operationCtx`, so it fires even when the parent context is already canceled. That's the callback-context question I raised separately, not a missing loop check — so I'd resolve it by deciding the ctx for `AfterMigration` rather than adding a guard here.

dkropachev: No extra check needed here. Every completed statement is recorded before EOF, and the final AwaitSchemaAgreement(ctx) still returns cancellation.

migrate/migrate.go
dkropachev: Kept detachment for a started statement and its progress write. Documented the deadline behavior and added deadline regression coverage.

migrate/migrate.go
dkropachev: Updated the context documentation to describe checks between statements and lifecycle callback behavior.

migrate/migrate.go
dkropachev: Changed BeforeMigration and AfterMigration to use the parent context. CallComment remains detached because it shares statement progress bookkeeping. Added docs and coverage.

migrate/migrate.go
dkropachev: Cancellation now reports the statement index and wraps the original error.

migrate/migrate.go
dkropachev: Changed the per-statement schema agreement wait to use the parent context and added a cancellation check after BeforeMigration.

migrate/migrate_test.go
dkropachev: Changed the wrappers to %w. Cancellation tests now use errors.Is.

migrate/migrate_test.go
```diff
@@ -379,6 +383,291 @@ func TestMigrationCallback(t *testing.T) {
...
+		t.Fatalf("after migration callback calls=%d expected 1", afterCalls)
+	}
+
+	if err := migrate.FromFS(context.Background(), session, f); err != nil {
+		t.Fatalf("resume migration: %v", err)
+	}
+	if afterCalls != 1 {
+		t.Fatalf("after migration callback calls after resume=%d expected 1", afterCalls)
+	}
+}
```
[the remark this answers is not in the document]
dkropachev: Addressed in 3e4b883: both tests now obtain cancellation from the migration loop, with boundary-specific assertions.

migrate/migrate.go
```diff
@@ -143,13 +143,17 @@ func Migrate(ctx context.Context, session gocqlx.Session, dir string) error {
 // FromFS executes new CQL files from a file system abstraction (io/fs.FS).
 // The provided FS has to be a flat directory containing *.cql files.
 //
+// Cancellation and deadlines stop migration between statements. Once a
+// statement starts, its execution, progress update, and final AfterMigration
+// callback finish before the cancellation error is returned.
```
nikagra: `AfterMigration` doesn't run on a mid-file cancel: the gate at L284 returns before the `i > done` block at L339, so it only fires when the file completes. That's why `TestMigrationFinalCallbackFinishesAfterCancellation` observes it (`0.cql` is a single statement) but `TestMigrationRecordsProgressAfterContextCanceled` doesn't.

Suggest: "its execution and progress update finish before the cancellation error is returned. `AfterMigration` runs only when the file's last statement completes."

migrate/migrate.go
```diff
@@ -319,8 +337,8 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 	}
 
 	if Callback != nil && i > done {
-		if err := Callback(ctx, session, AfterMigration, info.Name); err != nil {
-			return fmt.Errorf("after migration callback: %s", err)
+		if err := Callback(operationCtx, session, AfterMigration, info.Name); err != nil {
```
nikagra: Detached again here — an earlier reply said `BeforeMigration` and `AfterMigration` would both take the parent ctx, and `callback.go` now documents `AfterMigration` as detached, so I read the revert as deliberate.

Still worth reconsidering: the atomicity argument for detaching `CallComment` doesn't transfer. `CallComment` advances `done`, so the callback and its progress write must not be split. `AfterMigration` runs after the loop with no `update` following it, so detaching buys no atomicity — it only removes the caller's last cancellation handle on what is often the slowest callback.

migrate/migrate.go
```diff
@@ -295,12 +313,12 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
...
-				return fmt.Errorf("callback %s: %s", cb, err)
+			if err := Callback(operationCtx, session, CallComment, cb); err != nil {
+				return fmt.Errorf("callback %s: %w", cb, err)
 			}
 		} else if stmt != "" && !isComment(stmt) {
 			// Execute SQL statements (skip empty statements and comments)
-			q := session.ContextQuery(ctx, stmt, nil).RetryPolicy(nil)
+			q := session.ContextQuery(operationCtx, stmt, nil).RetryPolicy(nil)
 			if err := q.ExecRelease(); err != nil {
 				return fmt.Errorf("statement %d: %s", i, err)
```
nikagra: Missed `%w` here, and at L332. These two carry the actual driver errors (`ErrTimeoutNoResponse`, `RequestError`), so they're the ones most worth matching with `errors.Is`/`errors.As`.

migrate/migrate.go
```diff
@@ -274,15 +281,26 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 			continue
 		}
 
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("canceled before statement %d: %w", i, err)
+		}
```
nikagra, line 288: "canceled" is misleading when the cause is a deadline — the `DeadlineExceeded` tests produce `canceled before statement 3: context deadline exceeded`. `context ended before statement %d` covers both.

migrate/migrate_test.go
```diff
@@ -379,6 +383,301 @@ func TestMigrationCallback(t *testing.T) {
...
+	select {
+	case <-c.done:
+		return context.DeadlineExceeded
+	default:
+		return nil
+	}
+}
+
+func (c *manualDeadlineContext) expire() {
+	close(c.done)
```
nikagra: `close` isn't idempotent, so this helper can't be called twice: L541 wraps it in a `sync.Once` while L604 calls it bare, safe only because that `CALL` is skipped on resume. Moving the `Once` in here makes both call sites safe and lets `expireOnce` go.

migrate/migrate_test.go
```diff
@@ -379,6 +383,301 @@ func TestMigrationCallback(t *testing.T) {
...
+	defer session.Close()
+	recreateTables(t, session)
+
+	f := memfs.New()
+	err := f.WriteFile("0.cql", []byte(fmt.Sprintf(insertMigrate, 0)+";\n-- CALL cancel;\n-- CALL after_cancel;\n"+fmt.Sprintf(insertMigrate, 1)+";"), fs.ModePerm)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
nikagra, line 397: No `defer cancel()` here, or at L455/L495 — `cancel` is only reachable from inside the callback, so an early `t.Fatal` leaks the context.

migrate/migrate.go
```diff
@@ -274,15 +281,26 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
...
+
 		if DefaultAwaitSchemaAgreement.ShouldAwait(AwaitSchemaAgreementBeforeEachStatement) {
 			if err = session.AwaitSchemaAgreement(ctx); err != nil {
-				return fmt.Errorf("awaiting schema agreement: %s", err)
+				return fmt.Errorf("awaiting schema agreement: %w", err)
```
nikagra: Cancellation here loses the statement index that every other path now reports — cancel during this wait and you get `awaiting schema agreement: context canceled`, cancel a moment later and you get `canceled before statement N`. Same event, different diagnostics.

`awaiting schema agreement before statement %d: %w` would line them up.

migrate/migrate.go
```diff
@@ -274,15 +281,26 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 			continue
 		}
 
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("canceled before statement %d: %w", i, err)
+		}
+
 		if Callback != nil && i == 1 {
```
nikagra: Pre-existing, but this PR makes it matter more: `i == 1` is unreachable on resume, because `i <= done` skips those iterations before this check. So a resumed migration gets no `BeforeMigration` while still getting `AfterMigration` at L340.

Now that cancel-then-resume is a first-class flow, a hook that does setup in `BeforeMigration` silently won't run for the resumed half.

migrate/migrate_test.go
```diff
@@ -379,6 +383,301 @@ func TestMigrationCallback(t *testing.T) {
...
+	if !errors.Is(err, context.DeadlineExceeded) || !strings.Contains(err.Error(), "canceled before statement 2") {
+		t.Fatalf("FromFS() error=%v expected deadline exceeded before statement 2", err)
+	}
+	select {
+	case hasDeadline := <-deadlineObserved:
+		if hasDeadline {
+			t.Fatal("CQL statement context retained parent deadline")
+		}
+	default:
+		t.Fatal("CQL statement was not observed")
```
nikagra: The non-blocking read only works because gocql invokes `ObserveQuery` synchronously on the executing goroutine (`session.go:2111`, non-speculative default). A blocking receive with a short `time.After` asserts the same thing without depending on that.

migrate/migrate_test.go
```diff
@@ -379,6 +383,301 @@ func TestMigrationCallback(t *testing.T) {
...
+
+	if done := migrationDone(t, session, "0.cql"); done != 2 {
+		t.Fatalf("migration done=%d expected 2", done)
+	}
+	if count := countMigrations(t, session); count != 1 {
+		t.Fatalf("migration statements executed=%d expected 1", count)
+	}
+	if afterCancelCalls != 0 {
+		t.Fatalf("post-cancellation callback calls=%d expected 0", afterCancelCalls)
+	}
```
nikagra, line 437: Worth also asserting `AfterMigration` did not fire here. That's the case my note on the `FromFS` doc is about — a mid-file cancel returns before the `i > done` block — and nothing currently pins it, so a refactor could start firing it on cancel unnoticed.

migrate/migrate.go
dkropachev: Accepted and fixed.

migrate/migrate.go
dkropachev: Accepted and fixed.

migrate/migrate.go
dkropachev: Accepted and fixed.

migrate/migrate.go
dkropachev: Accepted and fixed.

migrate/migrate_test.go
dkropachev: Accepted and fixed.

migrate/migrate_test.go
dkropachev: Accepted and fixed.

migrate/migrate.go
dkropachev: Accepted and fixed.

migrate/migrate.go
dkropachev: Accepted and fixed.

migrate/migrate_test.go
dkropachev: Accepted and fixed.

migrate/migrate_test.go
dkropachev: Accepted and fixed.

migrate/migrate.go
```diff
@@ -320,7 +339,7 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
 
 	if Callback != nil && i > done {
 		if err := Callback(ctx, session, AfterMigration, info.Name); err != nil {
-			return fmt.Errorf("after migration callback: %s", err)
+			return fmt.Errorf("after migration callback: %w", err)
 		}
 	}
```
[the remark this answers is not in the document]
dkropachev: Documented the at-most-once behavior. Durable retries: #392.

migrate/callback.go
```diff
@@ -25,7 +25,12 @@ const (
 // If error is returned the migration is aborted.
 // BeforeMigration and AfterMigration are triggered before and after processing
```
nikagra: `i == done+1` makes `BeforeMigration` fire once per resume, while `AfterMigration` fires at most once ever — `TestMigrationRecordsProgressAfterContextCanceled` pins both (`beforeCalls` 1→2 on the same file, `afterMigrationCalls` 0→1). This line still says "before and after processing of each migration file," which now matches neither. Non-idempotent setup in `BeforeMigration` silently repeats on resume.

Not suggesting Copilot's revert to `i == 1` — that's the bug from 3703661077. The asymmetry is right, it just needs stating.

L28-29: worth naming cancellation, since that's how the at-most-once caveat actually gets reached.

migrate/migrate_test.go
```diff
@@ -379,6 +383,317 @@ func TestMigrationCallback(t *testing.T) {
...
+	cluster := gocqlxtest.CreateCluster()
+	cluster.QueryObserver = queryObserverFunc(func(queryCtx context.Context, query gocql.ObservedQuery) {
+		if !strings.Contains(query.Statement, "INSERT INTO gocqlx_test.migrate_table") {
+			return
+		}
+		_, hasDeadline := queryCtx.Deadline()
+		deadlineObserved <- hasDeadline
+		ctx.expire()
+	})
+	if err := gocqlxtest.CreateKeyspace(cluster, "gocqlx_test"); err != nil {
```
nikagra: `CreateKeyspace` opens with `DROP KEYSPACE IF EXISTS` (gocqlxtest.go:86), so this wipes `gocqlx_test` mid-run. Every other test goes through `createSessionFromCluster`, whose `initOnce` exists to make that drop happen once per binary.

Passes only on ordering — breaks under `-shuffle=on`, a `-run` subset, or a test inserted before :603. Also hard-codes the keyspace instead of `*flagKeyspace`.

Exporting `CreateSessionFromCluster(tb, cluster)` would give this test what it wanted: observer, no drop.

migrate/migrate.go
```diff
@@ -211,7 +216,7 @@ func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
 // allowing for resumption of partially completed migrations.
 //
 // Parameters:
-//   - ctx: context for cancellation and timeouts
+//   - ctx: context checked between statements and passed to BeforeMigration
```
nikagra: Also passed to `AfterMigration` (:341) and both schema-agreement waits (:259, :300).

migrate/migrate.go
```diff
@@ -274,15 +282,26 @@ func applyMigration(ctx context.Context, session gocqlx.Session, f fs.FS, path s
...
+		}
+
+		if Callback != nil && i == done+1 {
 			if err := Callback(ctx, session, BeforeMigration, info.Name); err != nil {
-				return fmt.Errorf("before migration callback: %s", err)
+				return fmt.Errorf("before migration callback: %w", err)
 			}
 		}
 
+		if err := ctx.Err(); err != nil {
```
nikagra: Nit: outside the `Callback` branch this duplicates :285 whenever `Callback == nil`. Moving it inside `if Callback != nil && i == done+1` scopes it to the case it exists for.

migrate/migrate_test.go
```diff
@@ -379,6 +383,317 @@ func TestMigrationCallback(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	afterCancelCalls := 0
+	beforeCalls := 0
+	afterMigrationCalls := 0
+	migrate.Callback = func(ctx context.Context, session gocqlx.Session, ev migrate.CallbackEvent, name string) error {
```
nikagra: Nit: `ctx` shadows the outer `ctx`, and the assertion at :411 depends on the shadow. The sibling tests use `callbackCtx`.

migrate/migrate_test.go
```diff
@@ -379,6 +383,317 @@ func TestMigrationCallback(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	afterCalls := 0
+	migrate.Callback = func(callbackCtx context.Context, session gocqlx.Session, ev migrate.CallbackEvent, name string) error {
+		switch ev {
+		case migrate.CallComment:
+			cancel()
+		case migrate.AfterMigration:
+			afterCalls++
+			if err := callbackCtx.Err(); err == nil {
```
nikagra: Nit: `else` after `return`. `revive` is enabled, so `make check` may fail on it.

nikagra (commented):
Prior round's ten points all verified fixed. Approve once the `CallbackFunc` doc line lands.

No inline anchor (outside the diff hunks): `migrate.go:181` `calculate checksum for %q: %s` is the last `%s` left in the file; `Pending` uses `%w` for the same message at :115.

Never executed by any test, since `DefaultAwaitSchemaAgreement` defaults to `Disabled`: `:258-262` and `:299-306`, including the new `awaiting schema agreement before statement %d` and the post-wait recheck. Also no multi-file cancellation case.

Correcting my earlier round: these tests *are* run in CI — `make test` uses `-tags all`, which matches the build tag.

migrate/migrate_test.go
```diff
@@ -389,6 +704,16 @@ func countMigrations(tb testing.TB, session gocqlx.Session) int {
 	return v
 }
 
+func migrationDone(tb testing.TB, session gocqlx.Session, name string) int {
+	tb.Helper()
+
+	var v int
+	if err := session.Query("SELECT done FROM gocqlx_migrate WHERE name = ?", nil).Bind(name).Get(&v); err != nil {
```
nikagra: Nit: unqualified `gocqlx_migrate`; `countMigrations` (:701) and `recreateTables` qualify with `gocqlx_test.`.

migrate/callback.go
dkropachev: Updated CallbackFunc docs to describe BeforeMigration on resume and cancellation-related AfterMigration behavior.

migrate/migrate_test.go
dkropachev: Switched test to shared keyspace lifecycle via CreateSessionFromCluster; it no longer drops keyspace mid-run.

migrate/migrate.go
dkropachev: Expanded ctx parameter docs to cover AfterMigration and both schema-agreement waits.

migrate/migrate.go
dkropachev: Scoped post-callback cancellation check inside callback branch.

migrate/migrate_test.go
dkropachev: Renamed callback parameter to callbackCtx.

migrate/migrate_test.go
dkropachev: Removed else and return saved context error directly.

migrate/migrate_test.go
dkropachev: Qualified query with gocqlx_test keyspace.

nikagra (approved):
Verified against 94b5a46 — everything from the earlier rounds is addressed.

Ran the integration suite against a local Scylla, which the PR description notes hadn't been possible and which CI's Build job doesn't cover: `go test -cpu 1 -count=1 -cover -race -tags all ./migrate` passes (16 tests, 62.6% coverage), including all five new ones. `-race` is clean, so the `ObserveQuery` callback in `TestMigrationCQLStatementUsesDetachedContext` is fine writing to `deadlineObserved` and calling `expire()` from the driver goroutine. `-shuffle=on` plus fixed seeds 1/7/42 all pass, which confirms `CreateSessionFromCluster` removed the ordering dependence on the keyspace drop. `make check` reports 0 issues.

One non-blocking leftover, in the body only because the line falls between diff hunks and has no inline anchor: `migrate.go:181` (`calculate checksum for %q: %s`) is the last `%s` remaining after this PR's `%w` sweep. Fine as a follow-up.

PR #393, opened 2026-08-04
udt: support automatic UDTs in maps

### LINKED ISSUE #188: gocqlx.UDT by map[string]UDT is null
udt can not marshal data.
use interface custom MarshalUDT, UnmarshalUDT worked.

dkropachev:
Closes #188.

## Summary

- support automatic `gocqlx.UDT` values and keys in maps
- recursively support mapper-aware UDTs in nested maps, lists, sets, slices, and arrays
- preserve the configured mapper and strict mode throughout nested codecs
- apply the same wrapping to batch bind paths without mutating caller arguments
- preserve `gocql.Marshaler`, `gocql.Unmarshaler`, `gocql.UDTMarshaler`, and `gocql.UDTUnmarshaler` precedence
- use read-only field traversal while marshaling, preserving nil fields and avoiding mutation or reflection panics for map and collection elements
- clear stale automatic UDT destinations when a top-level or nested CQL UDT is null

## Scope and ownership

The following behavior belongs in gocqlx and is implemented here:

- recognition of the `gocqlx.UDT` marker
- mapper-aware `db` field lookup and strict-mode handling
- read-only marshaling traversal
- preservation of existing gocql custom-codec precedence
- query and batch wrapping

Two narrow compatibility plugs remain internal while upstream gocql APIs and behavior are discussed:

1. **Recursive collection framing**
   
   `udtCodec` owns mapper-aware traversal of collection elements and currently duplicates private gocql collection framing. This is internal implementation detail; no new public framing API is exposed by gocqlx.
   
   Upstream request: https://github.com/scylladb/gocql/issues/985
   
   Removal condition: replace the custom framing after gocql provides a recursive collection-element codec hook and gocqlx raises its minimum dependency to that release.

2. **Null automatic UDT clearing**
   
   gocql currently leaves stale destination state when a wholly null UDT is decoded through `UDTUnmarshaler`. gocqlx handles this only for its own automatic UDT wrapper and only when `data == nil`, at both top-level and nested positions. User-defined `UDTUnmarshaler` implementations remain delegated to gocql unchanged.
   
   Upstream bug: https://github.com/scylladb/gocql/issues/986
   
   Removal condition: remove the targeted workaround after gocql defines and implements null handling for `UDTUnmarshaler` and gocqlx adopts that release.

Mapper-aware recursion inside CQL tuples and vectors is intentionally deferred to #395. Slices and arrays in this PR are Go representations of the supported CQL list and set paths.

General gocql collection extension APIs and null semantics for arbitrary custom UDT codecs are intentionally out of scope for this PR.

## Coverage

- concrete and pointer map values
- UDT map keys
- interface-typed UDT values
- nested maps, lists, sets, slices, and arrays
- strict mode on marshal and unmarshal
- nested UDT map fields
- nil, empty, and null-valued maps
- top-level and nested null UDT values
- custom gocql codec precedence
- caller argument preservation during batch binding
- collection framing parity with gocql
- live Scylla round trips for UDT map values and keys

## Testing

- `go test -race -count=1 . ./qb ./table ./migrate ./dbutil ./gocqlxtest`
- `go test -run "^$" -tags all ./...`
- `go vet ./...`
- `git diff --check`

`golangci-lint` was unavailable in the local environment; repository CI remains the lint authority.

### DIFF
--- a/batchx.go
+++ b/batchx.go
@@ -65,7 +65,7 @@ func (b *Batch) BindStruct(qry *Queryx, arg interface{}) error {
 	if err != nil {
 		return err
 	}
-	b.Query(qry.Statement(), args...)
+	b.Query(qry.Statement(), udtWrapSlice(qry.Mapper, qry.strict, args)...)
 	return nil
 }
 
@@ -75,7 +75,7 @@ func (b *Batch) Bind(qry *Queryx, args ...interface{}) error {
 	if len(qry.Names) != len(args) {
 		return fmt.Errorf("query requires %d arguments, but %d provided", len(qry.Names), len(args))
 	}
-	b.Query(qry.Statement(), args...)
+	b.Query(qry.Statement(), udtWrapSlice(qry.Mapper, qry.strict, args)...)
 	return nil
 }
 
@@ -86,7 +86,7 @@ func (b *Batch) BindMap(qry *Queryx, arg map[string]interface{}) error {
 	if err != nil {
 		return err
 	}
-	b.Query(qry.Statement(), args...)
+	b.Query(qry.Statement(), udtWrapSlice(qry.Mapper, qry.strict, args)...)
 	return nil
 }
 
@@ -97,7 +97,7 @@ func (b *Batch) BindStructMap(qry *Queryx, arg0 interface{}, arg1 map[string]int
 	if err != nil {
 		return err
 	}
-	b.Query(qry.Statement(), args...)
+	b.Query(qry.Statement(), udtWrapSlice(qry.Mapper, qry.strict, args)...)
 	return nil
 }
 
--- a/udt.go
+++ b/udt.go
@@ -5,8 +5,13 @@
 package gocqlx
 
 import (
+	"bytes"
+	"encoding/binary"
 	"fmt"
+	"math"
 	"reflect"
+	"runtime"
+	"sync"
 
 	"github.com/gocql/gocql"
 	"github.com/scylladb/go-reflectx"
@@ -21,10 +26,17 @@ type UDT interface {
 var (
 	_ gocql.UDTMarshaler   = udt{}
 	_ gocql.UDTUnmarshaler = udt{}
+	_ gocql.Marshaler      = udtCodec{}
+	_ gocql.Unmarshaler    = udtCodec{}
+
+	marshalerInterface    = reflect.TypeOf((*gocql.Marshaler)(nil)).Elem()
+	udtMarshalerInterface = reflect.TypeOf((*gocql.UDTMarshaler)(nil)).Elem()
+	udtTypeCache          sync.Map
 )
 
 type udt struct {
-	field  map[string]reflect.Value
+	fields *reflectx.StructMap
+	mapper *reflectx.Mapper
 	value  reflect.Value
 	strict bool
 }
@@ -32,15 +44,16 @@ type udt struct {
 func makeUDT(value reflect.Value, mapper *reflectx.Mapper, strict bool) udt {
 	return udt{
 		value:  value,
-		field:  mapper.FieldMap(value),
+		fields: mapper.TypeMap(reflect.Indirect(value).Type()),
+		mapper: mapper,
 		strict: strict,
 	}
 }
 
 func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
-	value, ok := u.field[name]
+	value, ok := u.fieldByName(name, true)
 	if ok {
-		return gocql.Marshal(info, value.Interface())
+		return marshalUDTValue(info, value, u.mapper, u.strict)
 	}
 	if !u.strict {
 		return nil, nil
@@ -49,9 +62,9 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
 }
 
 func (u udt) UnmarshalUDT(name string, info gocql.TypeInfo, data []byte) error {
-	value, ok := u.field[name]
+	value, ok := u.fieldByName(name, false)
 	if ok {
-		return gocql.Unmarshal(info, data, value.Addr().Interface())
+		return unmarshalUDTValue(info, data, value.Addr(), u.mapper, u.strict)
 	}
 	if !u.strict {
 		return nil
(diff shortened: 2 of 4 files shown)


udt.go
```diff
@@ -51,28 +58,267 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+		return fmt.Errorf("cannot unmarshal %s into an invalid value", info)
+	}
+	if value.CanInterface() {
+		if _, ok := value.Interface().(gocql.Unmarshaler); ok {
+			return gocql.Unmarshal(info, data, value.Interface())
+		}
+	}
+	if value.Type().Implements(autoUDTInterface) {
+		return gocql.Unmarshal(info, data, makeUDT(value, mapper, strict))
+	}
```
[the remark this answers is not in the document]
dkropachev: Both directions now preserve custom codec precedence: gocql.Marshaler/gocql.Unmarshaler and the UDT-specific codec interfaces are checked before automatic mapping. This also keeps embedded codecs such as FullNameUDT on the same path for reads and writes.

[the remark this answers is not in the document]
dkropachev: Both directions now use the same precedence: custom gocql codecs first, followed by automatic UDT mapping. The precedence coverage exercises both marshal and unmarshal paths.

udt.go
```diff
@@ -51,28 +58,267 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+
+func unmarshalUDTMap(info gocql.TypeInfo, data []byte, value reflect.Value, mapper *reflectx.Mapper, strict bool) error {
+	mapInfo, ok := info.(gocql.CollectionType)
+	if !ok || value.Kind() != reflect.Map {
+		return gocql.Unmarshal(info, data, value.Addr().Interface())
+	}
+	if data == nil {
+		value.Set(reflect.Zero(value.Type()))
+		return nil
+	}
```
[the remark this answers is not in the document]
dkropachev: The caller now rejects non-settable collection destinations before dispatching to the map decoder. Map, slice, and array non-pointer cases return errors instead of panicking.

udt.go
```diff
@@ -51,28 +58,267 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+
+	n, rest, err := readUDTSize(data)
+	if err != nil {
+		return err
+	}
+	if n < 0 {
+		return fmt.Errorf("negative map size %d", n)
+	}
+	value.Set(reflect.MakeMapWithSize(value.Type(), n))
+	for i := 0; i < n; i++ {
```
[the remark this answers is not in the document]
dkropachev: This decoder intentionally follows gocql collection framing and allocation behavior so UDT collections remain compatible with ordinary collections. A separate bound here would diverge from the upstream map decoder; that hardening should be applied upstream and inherited here.

udt.go
```diff
@@ -51,28 +58,267 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	}
+	seen[t] = true
+	switch t.Kind() {
+	case reflect.Ptr:
+		return containsUDT(t.Elem(), seen)
+	case reflect.Map:
+		return containsUDT(t.Key(), seen) || containsUDT(t.Elem(), seen)
+	}
+	return false
+}
```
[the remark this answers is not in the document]
dkropachev: Detection and framing now cover slices, arrays, lists, sets, map-backed sets, and nested collection combinations. Coverage includes map-of-list, map-of-array, list-of-map, and both set representations; automatic UDT fields recurse through the same path.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+		return nil
+	}
+	value = indirectInterface(value)
+	if isNil(value) {
+		return value.Interface()
+	}
 	if value.Type().Implements(autoUDTInterface) {
 		return makeUDT(value, mapper, strict)
 	}
+	if containsUDT(value.Type(), make(map[reflect.Type]bool)) {
```
nikagra: `iterx.go:305` (`fieldsByTraversal`) calls `udtWrapValue` for every column of every row of every `StructScan`/`Get`/`Select`, so each non-UDT field now pays a map allocation plus a type-graph walk where it used to pay one type-assert. That regresses users with no UDTs at all. `containsUDT` is pure in `t`, so a package-level `sync.Map` memo fixes it. Could you add before/after `benchmark_test.go` numbers?

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	if seen[t] {
+		return false
+	}
+	seen[t] = true
+	switch t.Kind() {
+	case reflect.Ptr, reflect.Slice, reflect.Array:
+		return containsUDT(t.Elem(), seen)
+	case reflect.Map:
+		return containsUDT(t.Key(), seen) || containsUDT(t.Elem(), seen)
+	case reflect.Interface:
```
nikagra, line 502: The reasoning is right, but this pulls `map[string]interface{}`, `[]interface{}` and `*interface{}` onto the new framing even when no UDT is involved. I traced the common shapes and they do fall back to `gocql.Unmarshal` correctly — but that argues for a test rather than leaving it implicit: assert `map[string]string`, `[]int` and `time.Time` stay unwrapped.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+			if _, ok := v.(gocql.UDTUnmarshaler); ok {
+				return gocql.Unmarshal(info, data, v)
+			}
+		}
+	}
+	if value.Type().Implements(autoUDTInterface) {
+		// TODO: remove this automatic-UDT workaround after
+		// https://github.com/scylladb/gocql/issues/986 is fixed upstream.
+		if data == nil {
+			return zeroUDTValue(value)
```
nikagra, line 279: This only fires for nested values. A top-level null UDT column still goes through `udt{}`, which implements `UDTUnmarshaler` but not `Unmarshaler`, so gocql's `len(data) == 0` early return (marshal.go:1630) leaves the destination stale — the same #986 bug in its most common form. `udtCodec` already handles it; returning it for `autoUDTInterface` types would give both levels one path. Otherwise worth calling out as out-of-scope, since the description reads as if #986 is covered.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+		item := reflect.New(value.Type().Elem())
+		if err := unmarshalUDTValue(mapInfo.Elem, valueData, item, mapper, strict); err != nil {
+			return err
+		}
+		value.SetMapIndex(key.Elem(), item.Elem())
+	}
+	return nil
+}
+
+func writeUDTBytes(buf *bytes.Buffer, data []byte) error {
```
nikagra: These reproduce gocql's private framing while the dependency is pinned by `replace`. A gocql bump that changed framing would corrupt the wire silently rather than fail the build. I checked and it matches v1.18.3 exactly today — asserting `marshalUDTMap` / `marshalUDTCollection` output is byte-identical to `gocql.Marshal` for a UDT-free `map[string]string` and `[]string` would keep it that way.

nikagra (commented):
Checked against `scylladb/gocql v1.18.3` `marshal.go` and `go-reflectx v1.0.1` rather than from memory: the framing matches gocql byte-for-byte, the nil-ptr-before-`Marshaler` ordering matches `gocql.Marshal`, and the read-only traversal fixes real mutation of the caller's struct via `FieldByIndexes`. Notes inline.

udt_test.go
```diff
@@ -0,0 +1,575 @@
...
+
+func (customUDTCodec) MarshalUDT(string, gocql.TypeInfo) ([]byte, error) {
+	return nil, errCustomUDTCodec
+}
+
+func (*customUDTCodec) UnmarshalUDT(string, gocql.TypeInfo, []byte) error {
+	return errCustomUDTCodec
+}
+
+func TestUDTInMaps(t *testing.T) {
```
nikagra, line 117: All coverage here is in-process against synthetic `TypeInfo`, and `-run "^$" -tags all` only proves compilation. For a wire-format change, one round-trip through `gocqlxtest` — `map<text, frozen<udt>>` with a UDT value, a UDT key and a null entry — would cover what unit tests structurally can't: that Scylla accepts these bytes.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	if value.Kind() == reflect.Ptr {
+		return marshalUDTValue(info, value.Elem(), mapper, strict)
+	}
+
+	switch info.Type() {
+	case gocql.TypeList, gocql.TypeSet:
+		return marshalUDTCollection(info, value, mapper, strict)
+	case gocql.TypeMap:
+		return marshalUDTMap(info, value, mapper, strict)
+	default:
```
nikagra, line 173: `TypeTuple` and vector fall through here, so their UDT elements revert to gocql's `cql` tags. The description promises "nested maps, lists, sets, slices, and arrays" without naming the exclusions — worth a TODO here and a line in the scope section.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	}
+	_, err := buf.Write(data)
+	return err
+}
+
+func writeUDTSize(buf *bytes.Buffer, size int) error {
+	if size > math.MaxInt32 {
+		return fmt.Errorf("collection too large")
+	}
+	return binary.Write(buf, binary.BigEndian, int32(size))
```
nikagra: `binary.Write` is reflection-based and runs once per collection plus twice per element. gocql writes the four bytes directly (marshal.go:689-692); `binary.BigEndian.PutUint32` would match it.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	if value.CanAddr() {
+		return gocql.Unmarshal(info, data, value.Addr().Interface())
+	}
+	return gocql.Unmarshal(info, data, value.Interface())
+}
+
+func unmarshalUDTMap(info gocql.TypeInfo, data []byte, value reflect.Value, mapper *reflectx.Mapper, strict bool) error {
+	mapInfo, ok := info.(gocql.CollectionType)
+	if !ok || value.Kind() != reflect.Map {
+		return gocql.Unmarshal(info, data, value.Addr().Interface())
```
nikagra: `unmarshalUDTCollection` routes this same fallback through `unmarshalUDTFallback`, which guards `CanAddr()`. Safe here only because the caller checks `CanSet()`, but that invariant isn't visible from this line.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	value := reflect.Indirect(u.value)
+	field, ok := u.fields.Names[name]
+	if !ok {
+		return reflect.Value{}, false
+	}
+	if !readOnly {
+		return reflectx.FieldByIndexes(value, field.Index), true
+	}
+
+	for _, index := range field.Index {
```
nikagra, line 87: This is `reflectx.FieldByIndexesReadOnly` plus a nil-intermediate-pointer guard the upstream helper lacks. Worth a comment saying that's why it's open-coded, or upstreaming the guard to go-reflectx.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+		return writeUDTSize(buf, -1)
+	}
+	if err := writeUDTSize(buf, len(data)); err != nil {
+		return err
+	}
+	_, err := buf.Write(data)
+	return err
+}
+
+func writeUDTSize(buf *bytes.Buffer, size int) error {
```
nikagra: Naming nit: these encode collection framing, not UDT framing, so the `UDT` prefix reads as the opposite of what they do. `writeCollectionSize` etc. would mirror the gocql function each one shadows.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	return unmarshalUDTValue(info, data, u.value, u.mapper, u.strict)
+}
+
 // udtWrapValue adds UDT wrapper if needed.
 func udtWrapValue(value reflect.Value, mapper *reflectx.Mapper, strict bool) interface{} {
+	if !value.IsValid() {
+		return nil
+	}
+	value = indirectInterface(value)
+	if isNil(value) {
```
nikagra, line 120: Worth noting this guard fixes a panic the description doesn't mention: the old `udtWrapSlice` asserted a typed-nil `(*SomeUDT)(nil)` to `UDT`, then `FieldMap` → `mustBe(Invalid, Struct)` panicked. A two-line regression test would stop it being refactored away.

udt.go
```diff
@@ -49,30 +56,457 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	case reflect.Ptr, reflect.Slice, reflect.Array:
+		return containsUDT(t.Elem(), seen)
+	case reflect.Map:
+		return containsUDT(t.Key(), seen) || containsUDT(t.Elem(), seen)
+	case reflect.Interface:
+		// The dynamic value may be an automatic UDT. Use the mapper-aware
+		// collection codec because the interface's method set cannot tell us.
+		return true
+	}
+	return false
```
nikagra, line 509: Struct falls through to false, so `map[string]Outer` — where `Outer` has a `db`-tagged UDT field but doesn't itself embed `gocqlx.UDT` — isn't wrapped. Consistent with the marker contract, but silent; a comment would save the next reader the trace.

batchx.go
```diff
@@ -75,7 +75,7 @@ func (b *Batch) Bind(qry *Queryx, args ...interface{}) error {
 	if len(qry.Names) != len(args) {
 		return fmt.Errorf("query requires %d arguments, but %d provided", len(qry.Names), len(args))
 	}
-	b.Query(qry.Statement(), args...)
+	b.Query(qry.Statement(), udtWrapSlice(qry.Mapper, qry.strict, args)...)
```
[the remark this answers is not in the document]
dkropachev: udtWrapSlice now wraps into a new slice. Regression coverage verifies the caller arguments stay unchanged while the batch receives the wrapped value.

udt.go
dkropachev: Type detection is now memoized by reflect.Type. The cached scalar wrapper path measures about 20.6 ns/op versus 48–67 ns/op for uncached detection, with zero allocations in both cases.

udt.go
dkropachev: Added coverage confirming map[string]string, []int, and time.Time remain unwrapped, plus interface-typed map/list framing parity tests.

udt.go
dkropachev: Top-level automatic UDTs now use udtCodec as well, so null data clears stale destinations. Added top-level regression coverage and updated the scope description.

udt.go
dkropachev: Added byte-for-byte framing parity coverage against gocql for UDT-free map and list values.

udt_test.go
dkropachev: Added a tagged gocqlxtest round trip through Scylla for UDT map values and frozen UDT keys. Null collection elements are rejected by Scylla, so that branch remains covered by the synthetic unit test.

udt.go
dkropachev: Tuple and vector recursion is now explicitly out of scope in the description and code TODO, tracked by #395.

udt.go
dkropachev: Collection sizes now use binary.BigEndian.PutUint32 before writing the bytes.

udt.go
dkropachev: The map fallback now routes through unmarshalUDTFallback, keeping the addressability guard local.

udt.go
dkropachev: Added a comment explaining the open-coded traversal and its nil-intermediate-pointer guard.

udt.go
dkropachev: Renamed the framing helpers to writeCollectionSize, writeCollectionBytes, readCollectionSize, and readCollectionBytes.

udt.go
dkropachev: Added typed-nil regression coverage to preserve the panic guard.

udt.go
dkropachev: Added a comment documenting that struct fields are intentionally not traversed unless the struct implements the UDT marker contract.

udt.go
```diff
@@ -49,30 +62,521 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+		}
+		value.Set(reflect.Zero(value.Type()))
+		return nil
+	}
+
+	n, rest, err := readCollectionSize(data)
+	if err != nil {
+		return err
+	}
+	if n < 0 {
```
sylwiaszunejko, line 359: maybe it would be also worth it to verify upper bound. Every element carries a 4-byte length prefix, so `n` elements require at least `4*n`
bytes remaining (`8*n` for a map, one prefix each for key and value). A 4-byte payload declaring 2^20 elements and carrying
zero bytes of element data results in a runtime `throw`, not a `panic` — no `recover()` in the driver or in user code can
contain it, and the whole process dies.

udt.go
```diff
@@ -49,30 +62,521 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	if isNil(value) {
+		return value.Interface()
+	}
 	if value.Type().Implements(autoUDTInterface) {
-		return makeUDT(value, mapper, strict)
+		return udtCodec{value: value, mapper: mapper, strict: strict}
+	}
+	if typeContainsUDT(value.Type()) {
+		return udtCodec{value: value, mapper: mapper, strict: strict}
 	}
```
sylwiaszunejko, line 128: `typeContainsUDT` calls `containsUDT` which opens with exactly `if t.Implements(autoUDTInterface) { return true }`
([udt.go:490](udt.go#L490)), so the second branch subsumes the first and both return the
identical value. The first branch is only a cache-lookup fast path, and it reads as if the
two cases differ.

udt.go
```diff
@@ -49,30 +62,521 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+	}
+	result := containsUDT(t, make(map[reflect.Type]bool))
+	udtTypeCache.Store(t, result)
+	return result
+}
+
+// hasPromotedCodec reports whether the codec implemented by t is promoted from
+// an anonymous field. Automatic UDT mapping takes precedence in that case so
+// an embedded field's codec cannot accidentally serialize the whole outer UDT.
+func hasPromotedCodec(t, codec reflect.Type) bool {
```
sylwiaszunejko, line 524: could we cache it? computing it on every requests seems expensive

udt.go
```diff
@@ -49,30 +62,521 @@ func (u udt) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
...
+		return unmarshalUDTCollection(info, data, value, mapper, strict)
+	case gocql.TypeMap:
+		if !value.CanSet() {
+			return fmt.Errorf("cannot unmarshal %s into non-pointer %s", info, value.Type())
+		}
+		return unmarshalUDTMap(info, data, value, mapper, strict)
+	default:
+		if value.CanAddr() {
+			return gocql.Unmarshal(info, data, value.Addr().Interface())
+		}
```
sylwiaszunejko, line 317: `unmarshalUDTFallback` exists ([udt.go:397-402](udt.go#L397-L402)) but the `default:` branch
  of `unmarshalUDTValue` ([udt.go:314-319](udt.go#L314-L319)) re-implements it inline,
  character for character.
Call `unmarshalUDTFallback` from the `default:` branch, and add the symmetric
`marshalUDTFallback(info, value)` for the five marshal sites.

PR #396, opened 2026-08-05
migrate: expose structured migration errors

dkropachev:
Follow-up to #383.

- expose sentinels for simple migration states
- expose typed errors for migration-name and checksum mismatches
- preserve filesystem errors through `fileChecksum`, `Pending`, and `FromFS`
- update callers and tests to use `errors.Is`/`errors.As`

Validation:
- `go test ./...`
- `go test -cpu 1 -count=1 -cover -race -tags all ./migrate` (78.1% coverage)
- `make check`

### DIFF
--- a/migrate/checksum.go
+++ b/migrate/checksum.go
@@ -21,7 +21,7 @@ func checksum(b []byte) string {
 func fileChecksum(f fs.FS, path string) (string, error) {
 	file, err := f.Open(path)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 	defer func() {
 		_ = file.Close()
--- a/migrate/errors.go
+++ b/migrate/errors.go
@@ -1,0 +1,135 @@
+// Copyright (C) 2026 ScyllaDB
+// Use of this source code is governed by a ALv2-style
+// license that can be found in the LICENSE file.
+
+package migrate
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrNoMigrations indicates that the migration file system contains no CQL files.
+	ErrNoMigrations = errors.New("no migration files found")
+	// ErrDatabaseAhead indicates that the database contains more applied migrations
+	// than the migration file system.
+	ErrDatabaseAhead = errors.New("database is ahead")
+	// ErrInconsistentMigrations indicates that an applied migration does not match
+	// the migration at the same position in the file system.
+	ErrInconsistentMigrations = errors.New("inconsistent migrations")
+	// ErrChecksumMismatch indicates that an applied migration file's checksum has changed.
+	ErrChecksumMismatch = errors.New("migration checksum mismatch")
+	// ErrNoMigrationStatements indicates that a migration file contains no statements.
+	ErrNoMigrationStatements = errors.New("no migration statements found")
+	// ErrMissingCallbackHandler indicates that a CALL comment has no registered handler.
+	ErrMissingCallbackHandler = errors.New("missing callback handler")
+)
+
+// NoMigrationsError describes the file pattern for which no migrations were found.
+type NoMigrationsError struct {
+	Pattern string
+}
+
+func (e *NoMigrationsError) Error() string {
+	return fmt.Sprintf("no migration files found matching %q", e.Pattern)
+}
+
+// Unwrap allows errors.Is to match ErrNoMigrations.
+func (e *NoMigrationsError) Unwrap() error {
+	return ErrNoMigrations
+}
+
+// DatabaseAheadError describes migration counts when the database contains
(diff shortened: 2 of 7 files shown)


migrate/migrate.go
```diff
@@ -178,7 +178,7 @@ func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
 		}
 		c, err := fileChecksum(f, fm[i])
 		if err != nil {
-			return fmt.Errorf("calculate checksum for %q: %s", fm[i], err)
+			return fmt.Errorf("calculate checksum for %q: %w", fm[i], err)
```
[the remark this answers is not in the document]
dkropachev: Fixed in 218ee70. fileChecksum now returns fs.FS.Open errors, both Pending and FromFS preserve them through %w, and unit plus live integration tests assert errors.Is against the underlying cause. The PR also exposes structured migration state errors and updates existing consumers to use errors.Is/errors.As.

migrate/errors.go
```diff
@@ -0,0 +1,59 @@
...
+// ChecksumMismatchError describes a migration file whose checksum differs
+// from the checksum recorded in the database.
+type ChecksumMismatchError struct {
+	Path     string
+	Expected string
+	Actual   string
+}
+
+func (e *ChecksumMismatchError) Error() string {
+	return fmt.Sprintf("file %q was tampered with, expected md5 %s", e.Path, e.Expected)
```
sylwiaszunejko: Why `Actual` is not printed here?

migrate/checksum.go
```diff
@@ -21,7 +21,7 @@ func checksum(b []byte) string {
 func fileChecksum(f fs.FS, path string) (string, error) {
 	file, err := f.Open(path)
 	if err != nil {
-		return "", nil
+		return "", err
```
sylwiaszunejko, line 24: This is not a cosmetic change and it is the most consequential line in the commit, yet
the subject line (`expose structured migration errors`) and the body say nothing about it. Either split it into its own commit, or extend the commit body.

migrate/migrate.go
```diff
@@ -163,25 +163,33 @@ func FromFS(ctx context.Context, session gocqlx.Session, f fs.FS) error {
...
 	if len(fm) == 0 {
-		return fmt.Errorf("no migration files found")
+		return ErrNoMigrations
 	}
 	sort.Strings(fm)
 
 	// verify migrations
 	if len(dbm) > len(fm) {
-		return fmt.Errorf("database is ahead")
+		return ErrDatabaseAhead
```
sylwiaszunejko: this is not a regression, but maybe it would be worth it to add more information to this error e.g. migrations applied vs available

migrate/migrate.go
sylwiaszunejko: The commit structured 5 of 7 programmatically-interesting error conditions in the
package. The two it skipped are the same condition reported twice, inconsistently:

```go
// migrate.go:323
return fmt.Errorf("statement %d: missing callback handler while trying to call %s", i, cb)

// callback.go:67
return errors.New("missing handler")
```

Both mean "a `-- CALL <name>;` comment was reached with no handler registered". That is a caller
*configuration* bug — exactly the class most worth detecting programmatically, since the fix is in
the caller's code. And the two messages don't even agree on wording, so a user string-matching
today has to match both.

migrate/checksum.go
dkropachev: Yeah, I even though to move it to a separate pr, let me fix it

migrate/errors.go
dkropachev: Fixed in ddb32d2. ChecksumMismatchError.Error now includes both expected and actual checksum values, with regression coverage for the complete message.

migrate/checksum.go
dkropachev: Updated in ddb32d2. I amended the commit body to explicitly document returning fs.FS.Open failures from fileChecksum and preserving the original cause through error wrapping.

migrate/migrate.go
dkropachev: Implemented in ddb32d2. FromFS now returns *DatabaseAheadError with Applied and Available fields, while errors.Is continues to match ErrDatabaseAhead. Unit and integration tests cover the counts.

migrate/migrate.go
dkropachev: Implemented in ddb32d2. Both missing-handler paths now return *MissingCallbackHandlerError with the callback name; errors.Is matches ErrMissingCallbackHandler and errors.As exposes structured context. Tests cover CallbackRegister and migration execution.

sylwiaszunejko (approved):

